PR #36, opened 2024-09-26
Upgrade preconf dependencies and fix all warnings

Akagi201:


crates/utils/src/signing.rs
```diff
@@ -40,36 +39,24 @@ pub fn verify_signed_builder_message<T: Merkleized>(
...
 ) -> Result<Root, Error> {
     let fork = context.fork_for(slot);
     let fork_version = context.fork_version_for(fork);
-    let domain = compute_domain(
-        DomainType::BeaconProposer,
-        Some(fork_version),
-        Some(*genesis_validators_root),
-        context,
-    )?;
+    let domain = compute_domain(DomainType::BeaconProposer, Some(fork_version), Some(*genesis_validators_root), context)?;
```
0w3n-d, line 50: The vast majority of this PR is stuff like this, where you take something that's been spread over multiple lines for legibility and collapse it to one line. What's the reasoning behind this? Does it remove warnings, or is it just your preferred style?

PR #41, opened 2024-10-03
feat(config): add allowed builders from config file

thedevbirb:
This PR enables adding allowed builders from the YAML configuration file using the `builders` field.

Currently adding new builders requires connecting to a the running postgres instance used by the relay and add new builders manually. While adding new builders happens rarely, when running the relay in a docker compose setup like a Kurtosis devnet this feature can be really helpful.

### DIFF
--- a/crates/database/src/types/documents.rs
+++ b/crates/database/src/types/documents.rs
@@ -12,7 +12,7 @@ use helix_common::{
     bid_submission::BidTrace,
     builder_info::BuilderInfo,
     simulator::BlockSimError,
-    SubmissionTrace,
+    BuilderConfig, SubmissionTrace,
 };
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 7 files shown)


crates/database/src/types/documents.rs
```diff
@@ -79,11 +79,7 @@ impl DemotionDocument {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize)]
-pub struct BuilderInfoDocument {
-    pub pub_key: BlsPublicKey,
-    pub builder_info: BuilderInfo,
-}
+pub type BuilderInfoDocument = BuilderConfig;
```
thedevbirb, line 81: Not a huge fan of the type aliasing done here but this greatly reduces the amount of changes and avoids circular dependencies with the `common` crate.

thedevbirb:
@gd-0 I see the CI is failing but I get the same errors on the `develop` branch. How do we handle it?

0w3n-d:
Sorry my latest changes to develop have created some conflicts here

thedevbirb:
> Sorry my latest changes to develop have created some conflicts here

No problem! I have solved the conflicts and rebased. I hope it's good now :)

PR #43, opened 2024-10-04
fix: switch `proposer_duties` table primary key to `slot_number`

thedevbirb:
As title says this PR changes the `PRIMARY_KEY` of the `proposer_duties` table from `public_key` to `slot_number`, and updates the logic of `set_proposer_duties` method of the `PostgresDatabaseService` accordingly.

The reason this change is needed is that, even if rare, it might occur that a validator public key is chosen twice for the next proposer duties which leads to the second record not being written on the database table because there is a conflict with the primary key. This edge case happens quite often in a devnet which has a very small validator set size compared to Mainnet.

The downstream consequence of it is that when processing a new head event inside `process_head_event` in the `ChainEventUpdater` (see below)
https://github.com/gattaca-com/helix/blob/3533496355837a4bb96b330d7e6c49408228ed7e/crates/housekeeper/src/chain_event_updater.rs#L151-L175
some proposer duties are missing which might lead to `next_duty` to being `None`. The latter is then read inside the `submit_block`/`submit_header` functions throwing the error `BuilderApiError::ProposerDutyNotFound` incorrectly.

### DIFF
--- a/crates/database/src/postgres/migrations/V21__proposer_duties_fix.sql
+++ b/crates/database/src/postgres/migrations/V21__proposer_duties_fix.sql
@@ -1,0 +1,2 @@
+ALTER TABLE proposer_duties DROP CONSTRAINT proposer_duties_pkey;
+ALTER TABLE proposer_duties ADD PRIMARY KEY (slot_number);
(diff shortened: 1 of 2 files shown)


0w3n-d (commented):
We can't modify this V1 file because it will cause an error when it tries to apply it to an existing relay. Instead it will need to modify the proposer_duties table. Can you add a new file to migrations folder with the next number and modify the table there?

thedevbirb:
> We can't modify this V1 file because it will cause an error when it tries to apply it to an existing relay. Instead it will need to modify the proposer_duties table. Can you add a new file to migrations folder with the next number and modify the table there?

Sorry about that! I've seen that on the `develop` branch the latest migration would be `V21` so I added that. I hope it's good now.

PR #44, opened 2024-10-10
Add postgres ssl_mode config

Akagi201:
ssl_mode config options: "prefer", "require", no config means none ssl_mode.


0w3n-d:
Nice!

PR #45, opened 2024-10-14
chore(docker): `local.Dockerfile`, `.dockerignore`

thedevbirb:
The current Dockerfile (https://github.com/gattaca-com/helix/blob/main/Dockerfile#L12-L45) isn't suitable for local development from external contributors given it expects some AWS secrets and that the command is run from a specific folder (with subfolders like `./repos/${REPO_NAME}`.

As such this PR introduces a `local.Dockerfile` intended to be used for local development from contributors and adds a `.dockerignore` file to avoid copying large and unused files and directories slowing down image building times.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.env*
+.git/
+.github/
+audits/
+images/
+**/target/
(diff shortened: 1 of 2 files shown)


0w3n-d:
Awesome! thanks

PR #49, opened 2024-11-07
feat: Constraints API v0

thedevbirb:
This PR introduces the [Constraints API](https://docs.boltprotocol.xyz/technical-docs/api/builder) in Helix.

### DIFF
--- a/crates/api/src/builder/api.rs
+++ b/crates/api/src/builder/api.rs
@@ -192,7 +192,7 @@ where
 
     /// This endpoint returns a list of signed constraints for a given `slot`.
     ///
-    /// Implements this API: <https://docs.boltprotocol.xyz/api/relay#constraints>
+    /// Implements this API: <https://docs.boltprotocol.xyz/technical-docs/api/relay#constraints>
     pub async fn constraints(
         Extension(api): Extension<Arc<BuilderApi<A, DB, S, G>>>,
         Query(slot): Query<SlotQuery>,
@@ -230,7 +230,7 @@ where
 
     /// This endpoint returns a stream of signed constraints for a given `slot`.
     ///
-    /// Implements this API: <https://docs.boltprotocol.xyz/api/relay#constraints-stream>
+    /// Implements this API: <https://docs.boltprotocol.xyz/technical-docs/api/relay#constraints-stream>
     pub async fn constraints_stream(
         Extension(api): Extension<Arc<BuilderApi<A, DB, S, G>>>,
     ) -> Sse<impl Stream<Item = Result<Event, BuilderApiError>>> {
@@ -260,7 +260,7 @@ where
     /// This endpoint returns the active delegations for the validator scheduled to propose
     /// at the provided `slot`. The delegations are returned as a list of BLS pubkeys.
     ///
-    /// Implements this API: <https://docs.boltprotocol.xyz/api/relay#delegations>
+    /// Implements this API: <https://docs.boltprotocol.xyz/technical-docs/api/relay#delegations>
     pub async fn delegations(
         Extension(api): Extension<Arc<BuilderApi<A, DB, S, G>>>,
         Query(slot): Query<SlotQuery>,
@@ -528,7 +528,7 @@ where
     /// 6. Simulates the block to validate the payment.
     /// 7. Saves the bid and inclusion proof to the auctioneer.
     ///
-    /// Implements this API: <https://docs.boltprotocol.xyz/api/relay#blocks_with_proofs>
+    /// Implements this API: <https://docs.boltprotocol.xyz/technical-docs/api/relay#blocks_with_proofs>
     pub async fn submit_block_with_proofs(
         Extension(api): Extension<Arc<BuilderApi<A, DB, S, G>>>,
         req: Request<Body>,
--- a/crates/api/src/builder/api.rs
+++ b/crates/api/src/builder/api.rs
@@ -7,27 +7,36 @@ use std::{
 
 use axum::{
     body::{to_bytes, Body},
-    extract::ws::{Message, WebSocket, WebSocketUpgrade},
+    extract::{
+        ws::{Message, WebSocket, WebSocketUpgrade},
+        Query,
+    },
     http::{Request, StatusCode},
-    response::{IntoResponse, Response},
+    response::{
+        sse::{Event, KeepAlive},
+        IntoResponse, Response, Sse,
+    },
     Extension, Json,
 };
 use ethereum_consensus::{
-    configs::mainnet::{CAPELLA_FORK_EPOCH, SECONDS_PER_SLOT},
+    configs::mainnet::CAPELLA_FORK_EPOCH,
     phase0::mainnet::SLOTS_PER_EPOCH,
     primitives::{BlsPublicKey, Hash32},
     ssz::{self, prelude::*},
 };
 use flate2::read::GzDecoder;
-use futures::StreamExt;
+use futures::{Stream, StreamExt};
 use hyper::HeaderMap;
+use reth_primitives::B256;
 use tokio::{
     sync::{
+        broadcast,
         mpsc::{self, error::SendError, Receiver, Sender},
         RwLock,
     },
     time::{self, Instant},
 };
+use tokio_stream::wrappers::BroadcastStream;
 use tracing::{debug, error, info, warn};
 use uuid::Uuid;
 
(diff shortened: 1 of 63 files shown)


crates/api/src/constraints/api.rs
```diff
@@ -0,0 +1,461 @@
...
+        db: Arc<DB>,
+        chain_info: Arc<ChainInfo>,
+        constraints_handle: ConstraintsHandle,
+    ) -> Self {
+        Self { auctioneer, db, chain_info, constraints_handle }
+    }
+
+    /// Handles the submission of batch of signed constraints.
+    ///
+    /// Implements this API: <https://docs.boltprotocol.xyz/api/builder#constraints>
```
merklefruit, line 78: broken link, update to https://docs.boltprotocol.xyz/technical-docs/api/builder#constraints

merklefruit: update to https://docs.boltprotocol.xyz/technical-docs/api/builder#constraints

crates/api/src/constraints/api.rs
```diff
@@ -0,0 +1,461 @@
...
+            request_duration_ns = trace.request_finish.saturating_sub(trace.receive),
+            "submit_constraints request finished",
+        );
+
+        Ok(StatusCode::OK)
+    }
+
+    /// Handles delegating constraint submission rights to another BLS key.
+    ///
+    /// Implements this API: <https://docs.boltprotocol.xyz/api/builder#delegate>
```
merklefruit: update to https://docs.boltprotocol.xyz/technical-docs/api/builder#delegate

crates/api/src/constraints/api.rs
```diff
@@ -0,0 +1,461 @@
...
+            request_duration_ns = trace.request_finish.saturating_sub(trace.receive),
+            "delegate request finished",
+        );
+
+        Ok(StatusCode::OK)
+    }
+
+    /// Handles revoking constraint submission rights from a BLS key.
+    ///
+    /// Implements this API: <https://docs.boltprotocol.xyz/api/builder#revoke>
```
merklefruit: update to https://docs.boltprotocol.xyz/technical-docs/api/builder#revoke

crates/api/src/proposer/api.rs
```diff
@@ -469,6 +469,141 @@ where
...
+    ///
+    /// This function accepts a slot number, parent hash and public_key.
+    /// 1. Validates that the request's slot is not older than the head slot.
+    /// 2. Validates the request timestamp to ensure it's not too late.
+    /// 3. Fetches the best bid for the given parameters from the auctioneer.
+    /// 4. Fetches the inclusion proof for the best bid.
+    ///
+    /// The function returns a JSON response containing the best bid and inclusion proofs if found.
+    ///
+    /// Implements this API: <https://docs.boltprotocol.xyz/api/builder#get_header_with_proofs>
```
merklefruit: update to https://docs.boltprotocol.xyz/technical-docs/api/builder#get_header_with_proofs

crates/common/src/config.rs
```diff
@@ -249,13 +282,35 @@ pub enum Route {
...
+    GetHeaderWithProofs,
+
+    // Constraints API: Relay <https://docs.boltprotocol.xyz/api/relay>
+    /// Reference: <https://docs.boltprotocol.xyz/api/relay#constraints>
+    GetBuilderConstraints,
+    /// Reference: <https://docs.boltprotocol.xyz/api/relay#constraints_stream>
+    GetBuilderConstraintsStream,
+    GetBuilderDelegations,
+    /// Reference: <https://docs.boltprotocol.xyz/api/relay#blocks_with_proofs>
+    SubmitBlockWithProofs,
```
merklefruit: all these links need to be updated

crates/common/src/bid_submission/submission.rs
```diff
@@ -353,4 +363,8 @@ pub struct SignedBidSubmissionDeneb {
     pub execution_payload: ExecutionPayload,
     pub blobs_bundle: BlobsBundle,
     pub signature: BlsSignature,
+    /// The Merkle proofs of inclusion as needed by the Constraints API.
+    /// Reference: <https://docs.boltprotocol.xyz/api/builder#get_header_with_proofs>
```
merklefruit: update to https://docs.boltprotocol.xyz/technical-docs/api/builder#get_header_with_proofs

merklefruit (changes requested):
Just tagging some broken docs links

crates/common/src/api/constraints_api.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub validator_pubkey: BlsPublicKey,
+    pub delegatee_pubkey: BlsPublicKey,
+}
+
+impl SignableBLS for DelegationMessage {
+    fn digest(&self) -> [u8; 32] {
+        let mut hasher = Sha256::new();
+        hasher.update([self.action]);
+        hasher.update(&self.validator_pubkey.to_vec());
+        hasher.update(&self.delegatee_pubkey.to_vec());
```
gd-0: `as_slice` can be used here to save the intermediate allocation.  (same with the `to_vec` lower in this file)
```suggestion
        hasher.update(self.validator_pubkey.as_slice());
        hasher.update(self.delegatee_pubkey.as_slice());
```

crates/common/src/proofs.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+impl TryFrom<SignedConstraints> for SignedConstraintsWithProofData {
+    type Error = ProofError;
+
+    fn try_from(value: SignedConstraints) -> Result<Self, ProofError> {
+        let mut transactions = Vec::with_capacity(value.message.transactions.len());
+        for transaction in value.message.transactions.to_vec().iter() {
+            let tx = PooledTransactionsElement::decode_enveloped(transaction.to_vec().into())
+                .map_err(|e| ProofError::DecodingFailed(e.to_string()))?;
```
gd-0: Can save an allocation here
```suggestion
        for transaction in value.message.transactions.iter() {
            let tx = PooledTransactionsElement::decode_enveloped(Bytes::copy_from_slice(transaction.as_slice()))
                .map_err(|e| ProofError::DecodingFailed(e.to_string()))?;
```

crates/api/src/constraints/api.rs
```diff
@@ -0,0 +1,464 @@
...
+        }
+
+        // PERF: can we avoid calling the db?
+        let maybe_validator_pubkey = api.db.get_proposer_duties().await?.iter().find_map(|d| {
+            if d.slot == first_constraints.slot {
+                Some(d.entry.registration.message.public_key.clone())
+            } else {
+                None
+            }
+        });
```
gd-0: we really shouldn't call the db here if possible. We run a single db instance in prod so we could be fetching from Tokyo -> FR every time this is called.

Ideally we should store proposer duties in redis in the housekeeper. then do `api.auctioneer.get_proposer_duties` instead

crates/api/src/builder/api.rs
```diff
@@ -398,6 +517,255 @@ where
...
+    /// 1. Receives the request and decodes the payload into a `SignedBidSubmission` object.
+    /// 2. Validates the builder and checks against the next proposer duty.
+    /// 3. Verifies the signature of the payload.
+    /// 4. Fetches the constraints for the slot and verifies the inclusion proofs.
+    /// 5. Runs further validations against the auctioneer.
+    /// 6. Simulates the block to validate the payment.
+    /// 7. Saves the bid and inclusion proof to the auctioneer.
+    ///
+    /// Implements this API: <https://docs.boltprotocol.xyz/technical-docs/api/relay#blocks_with_proofs>
+    pub async fn submit_block_with_proofs(
```
gd-0: As far as I can see this function is identical to `submit_block` except it might verify the proofs? can these two functions call a common fn that has a flag to verify proofs or not?

crates/api/src/constraints/api.rs
0w3n-d: Yeah in the other apis we store these in memory and keep them updated with the chain updater. Should just do the same here?

0w3n-d (approved):
My main issue with this PR is it seems to assume everything is talking to the same redis. But maybe we can fix this in a follow up PR.

crates/api/src/builder/api.rs
thedevbirb: You're right! Actually I was planning to keep the same underlying implementation for both endpoints, for the reasons mentioned in https://github.com/chainbound/helix/pull/28. What do you think? If you agree, I plan to merge that PR first so the changes needed to address your review will be done in the correct place.

gd-0: Yep sounds good to me!

PR #57, opened 2024-12-05
metrics

ltitanb:
add prometheus metrics:
- simulator
- redis
- api
- db

### DIFF
--- a/crates/api/src/gossiper/grpc_gossiper.rs
+++ b/crates/api/src/gossiper/grpc_gossiper.rs
@@ -25,10 +25,10 @@ use crate::{
 
 use super::types::broadcast_cancellation::BroadcastCancellationParams;
 
-const HEADER_ENDPOINT: &str = "header";
-const PAYLOAD_ENDPOINT: &str = "payload";
-const GET_PAYLOAD_ENDPOINT: &str = "get_payload";
-const CANCELLATION_ENDPOINT: &str = "cancel";
+const HEADER_ID: &str = "header";
+const PAYLOAD_ID: &str = "payload";
+const GET_PAYLOAD_ID: &str = "get_payload";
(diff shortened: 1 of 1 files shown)


crates/api/src/builder/simulator/optimistic_simulator.rs
```diff
@@ -157,6 +159,8 @@ impl<A: Auctioneer, DB: DatabaseService> BlockSimulator for OptimisticSimulator<
         request_id: Uuid,
     ) -> Result<bool, BlockSimError> {
         if self.should_process_optimistically(&request, builder_info).await {
+            SimulatorMetrics::sim_count(true);
```
gd-0, line 162: should this not be named `optimistic_sim_count` ? 
might be misunderstanding what this does

crates/api/src/gossiper/grpc_gossiper.rs
```diff
@@ -23,6 +25,11 @@ use crate::{
 
 use super::types::broadcast_cancellation::BroadcastCancellationParams;
 
+const HEADER_ENDPOINT: &str = "header";
+const PAYLOAD_ENDPOINT: &str = "payload";
+const GET_PAYLOAD_ENDPOINT: &str = "get_payload";
+const CANCELLATION_ENDPOINT: &str = "cancel";
```
gd-0: these are more IDs than endpoints imo

Owen66 (approved):

0w3n-d (approved):

crates/api/src/builder/simulator/optimistic_simulator.rs
ltitanb: that counts both types of simulations (true = optimistic)

PR #66, opened 2024-12-09
chore: fix genesis timestamp for devnet

zsluedem:
We are testing helix in the devnet(kurtosis devnet) which the genensis time is different from public network. The genesis delay needs to be considered otherwise the time is always wrong.

The pr is mainly to fix the problems by including genesis delay for other networks.


zsluedem:
@ltitanb Sorry for late update. It is fixed. Could you please check.

PR #79, opened 2025-03-08
bugfix: validator registry decoding issue

ckartik:
* There was an issue with ethers.rs in it's ability to decode certain values needed for the function call to `isValidatorOptedIn`- this code changes the implementation to manually encode/decode values for the function call.


crates/api/src/service.rs
```diff
@@ -98,7 +98,7 @@ impl ApiService {
             db.clone(),
             multi_beacon_client.clone(),
             auctioneer.clone(),
-            EthereumPrimevService::new(config.primev_config.unwrap()).await.unwrap(),
+            EthereumPrimevService::new(config.primev_config.clone().unwrap()).await.unwrap(),
```
0w3n-d, line 101: Can we make this optional and remove the unwraps? The relay shouldn't crash if the primev config is not present.
The relay should be able to run without it.

ckartik: Sure, let me make a new PR to add this.

PR #92, opened 2025-04-01
alloy-migration-2: housekeeper

ltitanb:
migrate most of housekeeper 
add new types crate + utils function to use in the transition


crates/common/src/eth/electra.rs
```diff
@@ -224,26 +222,26 @@ fn kzg_commitment_merkle_proof(
         .get_mut(0..std::mem::size_of::<usize>())
         .unwrap()
         .copy_from_slice(&length.to_le_bytes());
-    let length_root = H256::from_slice(length_bytes.as_slice());
+    let length_root = B256::from_slice(length_bytes.as_slice());
```
gd-0, line 225: for all the places where we are doing this, can't we just use:
```rust
B256::from(lengh_bytes)
``` 
? 

Think it should just give the B256 ownership of the lengh_bytes so saves a copy

ltitanb: yep in this first pass im just focusing on getting the types to compile, there's a lot of stuff that will be deleted later on (eg this function)

PR #93, opened 2025-04-04
alloy-migration-3: use lighthouse types

ltitanb:
- add a new `helix_types` crate with most beacon types from Lighthouse, replacing previous ones from ethereum_consensus and alloy_rpc_types_beacon
- all Bls types are also from Lighthouse, this works nicely with the beacon types although the bls types can be a bit annoying to use at times (eg. no `as_bytes()` on BlsPublicKey, because of this I had to change some types from `&[u8]` to `Vec<u8>` in the db
- This uses a fork of LH with some small changes, https://github.com/ltitanb/lighthouse/tree/helix-changes , will upstream those once finalized
- after this, `ethereum_consensus` is used in just a few places, which can be removed in a final PR
- there's still some functions that can/will be removed (eg. the old sig verification)
- temporarily removed the constraints API as discussed
- tests, mock impls, and tidyup still TODO


crates/api/src/builder/api.rs
```diff
@@ -2361,59 +2142,54 @@ fn sanity_check_block_submission(
...
-                    expected: Hash32::try_from(expected_withdrawals_root.as_ref()).unwrap(),
-                });
-            }
+        if *payload_withdrawals_root != *expected_withdrawals_root {
+            return Err(BuilderApiError::WithdrawalsRootMismatch {
+                got: payload_withdrawals_root,
+                expected: expected_withdrawals_root,
+            });
         }
     }
```
gd-0: i don't think we need the 2 different if cases for `payload.is_full_payload()` - can just combine into one

crates/api/src/builder/api.rs
```diff
@@ -1481,7 +1323,7 @@ where
     /// Returns: the bid submission in an Arc.
     async fn verify_submitted_block(
         &self,
-        mut payload: SignedBidSubmission,
+        payload: SignedBidSubmission,
```
gd-0, line 1323: comment for this fn should be:
```
    /// This function verifies:
    /// 1. Verifies the payload signature.
    /// 2. Simulates the submission
    ///
    /// Returns: the bid submission in an Arc.
```

crates/api/src/proposer/api.rs
```diff
@@ -714,8 +566,9 @@ where
                 }
             };
 
+        // TODO: remove unwrap
```
gd-0: should impl this TODO.

crates/api/src/proposer/api.rs
```diff
@@ -754,8 +607,16 @@ where
...
+        // let block_hash = signed_blinded_block
+        //     .message()
+        //     .body()
+        //     .execution_payload()
+        //     .unwrap()
+
+        //     .clone()
+
+        // FIXME
+        let block_hash = B256::ZERO;
```
gd-0: -

crates/types/src/clock.rs
```diff
@@ -0,0 +1,41 @@
+use std::time::Duration;
+
+pub use lh_slot_clock::{SlotClock as SlotClockTrait, SystemTimeSlotClock as SlotClock};
+
+pub const MAINNET_GENESIS_TIME: u64 = 1606824023;
+pub const SEPOLIA_GENESIS_TIME: u64 = 1655733600;
+pub const HOLESKY_GENESIS_TIME: u64 = 1695902400;
+pub const SECONDS_PER_SLOT: u64 = 12;
+pub const SLOTS_PER_EPOCH: u64 = 32;
```
gd-0: these must be in lh already. Can we take these from there?

crates/api/src/proposer/api.rs
```diff
@@ -1640,22 +1434,12 @@ fn calculate_slot_time_info(
         chain_info.genesis_time_in_secs + (slot * chain_info.seconds_per_slot);
     let ms_into_slot =
         (request_time / 1_000_000) as i64 - (slot_start_timestamp_in_secs * 1000) as i64;
-    let duration_until_slot_start = chain_info.clock.duration_until_slot(slot);
+    // TODO: check this
+    let duration_until_slot_start = chain_info.clock.duration_to_slot(slot.into()).unwrap();
```
gd-0: shouldn't unwrap here

gd-0: use of clock here looks good to me

crates/api/src/proposer/api.rs
```diff
@@ -1140,80 +984,57 @@ where
...
-                Ok(header) => header,
-                Err(err) => {
-                    error!(
-                        %err,
-                        "error converting execution payload to header",
-                    );
-                    return Err(err.into());
-                }
-            };
+            try_execution_header_from_payload(&local_versioned_payload.execution_payload);
```
gd-0: shouldn't be called `try_...` anymore

crates/api/src/proposer/api.rs
```diff
@@ -1140,80 +984,57 @@ where
...
 
         info!(
             local_header = ?local_header,
-            provided_header = ?provided_header,
-            provided_version = %provided_signed_blinded_block.version(),
+            // provided_header = ?provided_header,
+            // provided_version = %provided_signed_blinded_block.version(),
```
gd-0: lets add these back or remove entirely

crates/api/src/proposer/api.rs
```diff
@@ -1140,80 +984,57 @@ where
...
-            ExecutionPayloadHeader::Capella(local_header) => {
-                let provided_header =
-                    provided_header.capella().ok_or(ProposerApiError::PayloadTypeMismatch)?;
-                if local_header != *provided_header {
-                    return Err(ProposerApiError::BlindedBlockAndPayloadHeaderMismatch);
-                }
-            }
+            ExecutionPayloadHeader::Bellatrix(_) |
+            ExecutionPayloadHeader::Capella(_) |
+            ExecutionPayloadHeader::Fulu(_) => return Err(ProposerApiError::PayloadTypeMismatch),
```
gd-0: something like: `ProposerApiError::UnsupportedBeaconChainVersion` seems better here

crates/api/src/proposer/api.rs
```diff
@@ -1140,80 +984,57 @@ where
...
-                    .blobs_bundle
-                    .as_ref()
-                    .map(|bundle| &bundle.commitments)
-                    .ok_or(ProposerApiError::BlobKzgCommitmentsMismatch)?;
+                let local_kzg_commitments = &local_versioned_payload.blobs_bundle.commitments;
 
                 let provided_kzg_commitments = body
                     .blob_kzg_commitments()
-                    .ok_or(ProposerApiError::BlobKzgCommitmentsMismatch)?;
+                    .map_err(|_| ProposerApiError::BlobKzgCommitmentsMismatch)?;
```
gd-0: I think this electra check should also check the execution requests from the get_payload call. Lets discuss

gd-0: https://github.com/flashbots/mev-boost-relay/blob/17ed0547bfc90e589ae1d01555cb9156191a1a17/services/api/utils.go#L84-L145

Looks like mev-boost doesn't + looks like requests aren't even in the header so might be ok here

crates/api/src/builder/api.rs
ltitanb: done in d55a2e576491da187f8a59a49597de33416a4d24

crates/api/src/builder/api.rs
ltitanb: 0e78f6edd6b1e343ac61ac0f22fc664fbbe93c9c

crates/api/src/proposer/api.rs
ltitanb: 0e78f6edd6b1e343ac61ac0f22fc664fbbe93c9c

crates/api/src/proposer/api.rs
ltitanb: 755940c36f69fc4189c73b91847eafc8c474eb1d

crates/types/src/clock.rs
ltitanb: removed seconds per slot and slots per epoch, the genesis times are not in LH as they are loaded dynamically 58064a192b681651918287f6684066dd26e1b5d4

crates/api/src/proposer/api.rs
ltitanb: 2ebb77480ade5937c3afd862cb59026ae0efe7ab

crates/api/src/proposer/api.rs
ltitanb: 502d0040fb9f486e93988dd4f53dfd5de6a49914

crates/api/src/proposer/api.rs
ltitanb: 502d0040fb9f486e93988dd4f53dfd5de6a49914

local.Dockerfile
0w3n-d, line 1: Why was this deleted?

crates/api/src/gossiper/types/broadcast_header.rs
```diff
@@ -27,36 +23,33 @@ impl BroadcastHeaderParams {
...
     pub fn to_proto(&self) -> grpc::BroadcastHeaderParams {
         grpc::BroadcastHeaderParams {
             signed_builder_bid: serde_json::to_vec(&self.signed_builder_bid).unwrap(),
-            bid_trace: ssz::prelude::serialize(&self.bid_trace).unwrap(),
+            bid_trace: BidTrace::as_ssz_bytes(&self.bid_trace),
             slot: self.slot,
             parent_hash: self.parent_hash.to_vec(),
-            proposer_pub_key: self.proposer_pub_key.to_vec(),
-            builder_pub_key: self.builder_pub_key.to_vec(),
+            proposer_pub_key: self.proposer_pub_key.serialize().to_vec(),
```
0w3n-d, line 48: Why is this .serialize() step now needed?

0w3n-d (approved):
Looks good

local.Dockerfile
ltitanb: moved to `Dockerfile` in https://github.com/gattaca-com/helix/commit/78d50f154b40b9f13d09bfa43d24012e612dbdd3

crates/api/src/gossiper/types/broadcast_header.rs
ltitanb: it's the new lighthouse BlsPubkey type, that doesn't have `to_vec` and requires to get the raw compressed bytes first

PR #109, opened 2025-04-14
feat: remove beacon trait

ltitanb:
- removed beacon client traits and mock impls
- add mock beacon node for testing
- didnt change any logic in the beacon clients, although we should probably tidy that up at some point
- some housekeeper tests are now broken but that needs a refactor anywyas so we can fix those later

### DIFF
--- a/crates/common/src/task.rs
+++ b/crates/common/src/task.rs
@@ -1,12 +1,6 @@
-use std::{future::Future, sync::OnceLock};
+use std::future::Future;
 
-use tokio::{runtime::Handle, task::JoinHandle};
-
-static RUNTIME: OnceLock<Handle> = OnceLock::new();
-
-pub fn init_runtime() {
-    RUNTIME.get_or_init(build_tokio_runtime);
-}
+use tokio::task::JoinHandle;
 
 pub fn spawn<F>(file: &str, line: u32, future: F) -> JoinHandle<F::Output>
 where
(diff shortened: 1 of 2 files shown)


crates/common/src/task.rs
```diff
@@ -1,43 +1,22 @@
...
-    RUNTIME.get_or_init(build_tokio_runtime);
-}
+use tokio::task::JoinHandle;
 
 pub fn spawn<F>(file: &str, line: u32, future: F) -> JoinHandle<F::Output>
 where
     F: Future + Send + 'static,
     F::Output: Send + 'static,
 {
     let label = format!("{file}:{line}");
```
gd-0, line 10: not for this PR but we should remember to remove this format. Should use 2 labels in the metric tag

crates/housekeeper/src/housekeeper_tests.rs
```diff
@@ -24,33 +26,36 @@ use crate::{
...
-        state_validators_has_been_read.clone(),
-        proposer_duties_has_been_read.clone(),
-    );
+    let beacon_client = MultiBeaconClient::new(vec![beacon_node.beacon_client().into()]);
+    // let beacon_client = MultiBeaconClient::new(
+    //     subscribed_to_head_events.clone(),
+    //     chan_head_events_capacity.clone(),
+    //     state_validators_has_been_read.clone(),
+    //     proposer_duties_has_been_read.clone(),
+    // );
```
gd-0, line 44: should remove the comments

ltitanb: have left as a reminder for the housekeeper cleanup, will fix in a separate PR

PR #117, opened 2025-04-16
feat: metadata provider

0w3n-d:


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -1212,6 +1212,7 @@ impl DatabaseService for PostgresDatabaseService {
                 &(trace.simulation as i64),
                 &(trace.auctioneer_update as i64),
                 &(trace.request_finish as i64),
+                &(trace.metadata),
```
gd-0, line 1215: this doesn't seem very extendable if `metadata` is just a string. Maybe we can have an entirely diff table where the metadata provider returns the insert instructions? Then we can just reference that entry here.

bin/helix/src/main.rs
```diff
@@ -77,6 +73,7 @@ async fn run(config: RelayConfig, keypair: BlsKeypair) -> eyre::Result<()> {
         chain_info,
         relay_signing_context,
         beacon_client,
+        Arc::new(DefaultMetadataProvider{}) as Arc<dyn helix_common::metadata_provider::MetadataProvider>,
```
gd-0: why does this need to be `dyn` ? I think this could all be decided at compile time

crates/database/src/postgres/postgres_db_service.rs
0w3n-d: I was thinking this column would hold a json string, which would effectively turn it into a standard noSQL collection like in mongo. If we really needed it slit into columns we can just have a stored procedure in postgres that parses the json and populates another table.

bin/helix/src/main.rs
0w3n-d: I'm not sure what you mean. A dyn is how you define a trait object. What would the alternative look like? Do you just mean I don't need to explicitly cast it here?

0w3n-d: Updated

gd-0: we can also remove `dyn` in the extension i think

crates/database/src/postgres/postgres_db_service.rs
gd-0: how difficult is it to setup the stored procedure in PG? to me on the outside this definitely seems like the preferable option as it's much cleaner and faster for queries

0w3n-d: It's pretty trivial. We can just add extra column that get auto populated when data is inserted into the metadata column. Like this:

`ALTER TABLE validator_registrations
ADD COLUMN ip_address TEXT GENERATED ALWAYS AS (metadata ->> 'ip') STORED;`

bin/helix/src/main.rs
0w3n-d: If I remove it from the extension I get this error "trait objects must include the `dyn` keyword"

crates/database/src/postgres/postgres_db_service.rs
gd-0: ok perfect - seems like that's the way to go then?

bin/helix/src/main.rs
gd-0: you won't be able to just remove the dyn. you'll need to add a generic param to the api like we do for the simulator etc, then use that.
e.g.,
```rust
impl<A, DB, S, G, M> BuilderApi<A, DB, S, G, M>
where
    A: Auctioneer + 'static,
    ....
    M: MetadataProvider,


...
Arc<M>
```

crates/database/src/postgres/postgres_db_service.rs
0w3n-d: Yep. Sounds good to me.

bin/helix/src/main.rs
0w3n-d: Do you think all of that extra code bloat is worth the small runtime cost of resolving the dyn? Isn't it a pretty cheap little vtable look up?

gd-0: I think given that we are planning on removing a lot of the unecassary generics in the api e.g., simulator the code bloat for now is worth it. This is directly in the hot-path as well so it's worth spending extra time optimising. dyn also has more cost than just the vtable lookup, also removes any chance of fn inlining etc.

0w3n-d: Ok fair enough. I'll make the change. I like how clean and simple this code is but I suppose runtime performance is what matters most.

PR #122, opened 2025-04-21
cleanup functions

ltitanb:
- avoid cloning withdrawals for root
- remove error for not fallible gossip fns
- remove async from sync fns

### DIFF
--- a/crates/api/src/proposer/get_header.rs
+++ b/crates/api/src/proposer/get_header.rs
@@ -13,7 +13,6 @@ use helix_common::{
 use helix_database::DatabaseService;
 use helix_datastore::Auctioneer;
 use helix_types::BlsPublicKey;
-use serde_json::json;
 use tokio::time::sleep;
 use tracing::{debug, error, info, warn, Instrument};
 
--- a/crates/housekeeper/src/slot_info.rs
+++ b/crates/housekeeper/src/slot_info.rs
@@ -102,8 +102,7 @@ impl CurrentSlotInfo {
         );
 
         // Discard payload attributes if already known
-        let payload_attributes_key =
-            &(payload_attributes.parent_hash, payload_attributes.slot.into());
+        let payload_attributes_key = &(payload_attributes.parent_hash, payload_attributes.slot);
         let mut all_payload_attributes = self.payload_attributes.write();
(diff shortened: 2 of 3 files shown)


crates/api/src/proposer/get_header.rs
```diff
@@ -176,23 +176,20 @@ where
...
-                        }
                     });
                 }
 
-                info!(bid = %json!(bid), "delivering bid");
+                let bid = json!(bid);
+                info!(%bid, "delivering bid");
 
-                // Return header
                 Ok(axum::Json(bid))
```
gd-0: does this not change the behaviour of the return as it's `json!`'d?

crates/housekeeper/src/slot_info.rs
```diff
@@ -60,16 +60,17 @@ impl CurrentSlotInfo {
...
     pub(crate) fn handle_new_slot(&self, slot_update: SlotUpdate, chain_info: &ChainInfo) {
-        let epoch = slot_update.slot / chain_info.slots_per_epoch();
+        let epoch = slot_update.slot.epoch(chain_info.slots_per_epoch());
+
         info!(
-            epoch,
-            slot_head = slot_update.slot,
-            slot_start_next_epoch = (epoch + 1) * chain_info.slots_per_epoch(),
+            %epoch,
+            slot_head =% slot_update.slot,
```
gd-0:
```suggestion
            slot_head = %slot_update.slot,
```

crates/housekeeper/src/slot_info.rs
```diff
@@ -97,7 +98,7 @@ impl CurrentSlotInfo {
         }
 
         info!(
-            slot = payload_attributes.slot,
+            slot =% payload_attributes.slot,
```
gd-0:
```suggestion
            slot = %payload_attributes.slot,
```

crates/api/src/proposer/get_header.rs
ltitanb: before we encoding to json twice, one for the log and one for the return now we just do it once. Will make it more explicit removing the macro

ltitanb: https://github.com/gattaca-com/helix/pull/122/commits/b03eb742f6d915221f862a44612de542d9556376

PR #124, opened 2025-04-21
refactor housekeeper

ltitanb:
refactor housekeeper loop

### DIFF
--- a/crates/housekeeper/src/housekeeper.rs
+++ b/crates/housekeeper/src/housekeeper.rs
@@ -161,6 +161,7 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
 
     #[tracing::instrument(skip(self))]
     async fn process_new_slot(&self, head_slot: Slot, block_hash: Option<B256>) {
+        let start = Instant::now();
         if self.slots.head_already_seen(head_slot) {
             return;
         }
@@ -223,13 +224,13 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
                             );
 
                             slots.update_proposer_duties(head_slot);
-                            info!(duration = ?start.elapsed(), "proposer duties task completed");
                         }
 
                         Err(err) => {
                             error!(%err, "failed to update proposer duties");
                         }
                     }
+                    info!(duration = ?start.elapsed(), "proposer duties task completed");
                 }
                 .in_current_span(),
             ));
@@ -290,6 +291,7 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
 
         // wait for all tasks, this should take less than one slot
         join_all(tasks).await;
+        info!(duration = ?start.elapsed(), "processed new slot");
     }
 
     /// Refresh the list of known validators by querying the beacon client.


crates/housekeeper/src/housekeeper.rs
```diff
@@ -420,53 +309,28 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
...
-                return Err(HousekeeperError::BeaconClientError(err));
-            }
-        };
+        let proposer_duties = self.fetch_duties(epoch.as_u64()).await?;
 
         // Check if signed validator registrations exist for each proposer duty
-        let pub_keys: Vec<BlsPublicKey> =
-            proposer_duties.iter().map(|duty| duty.public_key.clone()).collect();
+        let pubkeys: Vec<&BlsPublicKey> =
+            proposer_duties.iter().map(|duty| &duty.public_key).collect();
```
gd-0, line 391: if we always iter through this then need to clone the pubkeys anyway why add the double indirection of adding the refs here?

crates/housekeeper/src/housekeeper.rs
```diff
@@ -475,69 +339,49 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
...
-        last_proposer_duty_distance >= PROPOSER_DUTIES_UPDATE_FREQ
+    fn should_update_duties(&self, head_slot: Slot) -> bool {
+        let last_updated = self.slots.proposer_duties();
+        head_slot.as_u64().saturating_sub(last_updated) >= PROPOSER_DUTIES_UPDATE_FREQ
     }
 
     /// Updates primev builders and validators using pre-fetched proposer duties
-    async fn primev_update_with_duties(
-        self: &SharedHousekeeper<DB, A>,
+    async fn maybe_primev_update_with_duties(
```
gd-0, line 429: just noticed in this fn - if a builder that is already registered with us also registers with primev this will completely overwrite their entry. We need to just add a new builder id not overwrite the entire thing.

@0w3n-d fyi

crates/housekeeper/src/housekeeper.rs
```diff
@@ -556,82 +400,38 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
...
-    ) -> Result<Vec<ProposerDuty>, BeaconClientError> {
-        // Fetch duties for current epoch
+    async fn fetch_duties(&self, epoch: u64) -> Result<Vec<ProposerDuty>, BeaconClientError> {
         let (_, mut proposer_duties) = self.beacon_client.get_proposer_duties(epoch).await?;
 
-        // Fetch duties for next epoch
         match self.beacon_client.get_proposer_duties(epoch + 1).await {
-            Ok((_, mut next_duties)) => proposer_duties.append(&mut next_duties),
-            Err(err) => error!(err = %err, "Error fetching next proposer duties"),
+            Ok((_, next_duties)) => proposer_duties.extend(next_duties),
```
gd-0: why have we switched from `append` to `extend`? `append` is preferable as it doesn't need to iter and can move all elements over in 1 go

crates/housekeeper/src/housekeeper.rs
```diff
@@ -28,331 +31,222 @@ const PROPOSER_DUTIES_UPDATE_FREQ: u64 = 1;
...
-                            error!("could not get current slot");
-                        }
-                    }
+            _ = timer.tick();
+
+            let head_slot = self.chain_info.current_slot();
+            if let Some(duration) = self.chain_info.duration_into_slot(head_slot) {
+                if duration > CUTOFF_TIME {
+                    self.process_new_slot(head_slot).await
                 }
```
gd-0: if we just use a ticker here might we miss cases that were previously handled where we get 2 updates for the same slot?

crates/housekeeper/src/housekeeper.rs
```diff
@@ -28,331 +31,222 @@ const PROPOSER_DUTIES_UPDATE_FREQ: u64 = 1;
...
-                                    Err(_) => {
-                                        debug!("Primev update already in progress, skipping");
-                                    }
-                                }
-                            });
+                        if let Err(err) =
+                            housekeeper.maybe_primev_update_with_duties(proposer_duties).await
+                        {
+                            error!(%err, "failed to update primev duties");
                         }
```
gd-0: should this not be spawned in its own task?

crates/housekeeper/src/housekeeper.rs
```diff
@@ -28,331 +31,222 @@ const PROPOSER_DUTIES_UPDATE_FREQ: u64 = 1;
...
-                    match self.chain_info.clock.now() {
-                        Some(slot) => self.process_new_slot(slot.as_u64()).await,
-                        None => {
-                            error!("could not get current slot");
-                        }
-                    }
+            _ = timer.tick();
+
+            let head_slot = self.chain_info.current_slot();
+            if let Some(duration) = self.chain_info.duration_into_slot(head_slot) {
```
0w3n-d: What is going on here? This is not refactoring. This is changing how this works. Instead of advancing the slot when the beacon node receives a new block it's using the fail safe on every slot? That's only for edge cases where the previous slot is missed.

crates/housekeeper/src/housekeeper.rs
0w3n-d: It doesn't overwrite anything. If builder is already registered it just adds primev to builder_ids

gd-0: ah then maybe we should change the naming for store_builder_info to update ?

crates/housekeeper/src/housekeeper.rs
ltitanb: I ran the timings with a holesky beacon and seeing this
```
2025-04-23T09:51:15.207112Z  INFO slot (event) 4124956 3.207078508s, slot (clock) 4124956 3.20707912s 
2025-04-23T09:51:28.048731Z  INFO slot (clock) 4124957 4.048715711s 
2025-04-23T09:51:38.044902Z  INFO slot (event) 4124958 2.044879965s, slot (clock) 4124958 2.044881077s 
2025-04-23T09:51:50.254855Z  INFO slot (event) 4124959 2.25484232s, slot (clock) 4124959 2.25484276s 
2025-04-23T09:52:04.048563Z  INFO slot (clock) 4124960 4.048547768s 
2025-04-23T09:52:14.908158Z  INFO slot (event) 4124961 2.908143249s, slot (clock) 4124961 2.908143971s 
2025-04-23T09:52:26.172076Z  INFO slot (event) 4124962 2.172063362s, slot (clock) 4124962 2.172064043s 
2025-04-23T09:52:40.048790Z  INFO slot (clock) 4124963 4.048770255s 
2025-04-23T09:52:49.931693Z  INFO slot (event) 4124964 1.931670984s, slot (clock) 4124964 1.931671666s 
```
is the main change that we would keep processing blocks for a few seconds more or is there something else I'm missing?

crates/housekeeper/src/housekeeper.rs
ltitanb: it avoids cloning all the pubkeys since the `get_validator_registrations_for_pub_keys` in postgres works without ownership as well so we avoid cloning all those pubkeys

crates/housekeeper/src/housekeeper.rs
ltitanb: sounds good will leave it for a separate PR

crates/housekeeper/src/housekeeper.rs
ltitanb: we still check if we have already processed the incoming slot (same as before)

crates/housekeeper/src/housekeeper.rs
gd-0: ah thought we cloned them in all places anyway. sounds good

crates/housekeeper/src/housekeeper.rs
0w3n-d: Yeah I suppose the main impact would just be needlessly processing blocks for the previous slot when the chain had already moved on. I'd need to double check the logic around how proposers are set to ensure it has no impact there too.

ltitanb: updated to use the head events again just for safety, just changed to a timeout vs select before

PR #130, opened 2025-04-23
handle 'share' header for v2 and v3 submissions

vladimir-ea:


crates/api/src/builder/api.rs
```diff
@@ -406,10 +406,24 @@ where
 
         debug!(timestamp_request_start = trace.receive,);
 
+        let is_gossip_enabled = headers
+            .get("share")
+            .and_then(|v| v.to_str().ok().map(|val| val.to_lowercase() != "none"))
+            .unwrap_or(true);
```
gd-0: lets deser into an enum with all fields so it's easier to extend in the future

crates/api/src/builder/api.rs
```diff
@@ -406,10 +406,24 @@ where
 
         debug!(timestamp_request_start = trace.receive,);
 
+        let is_gossip_enabled = headers
+            .get("share")
```
ltitanb: let's move this to a constant and give it a more descriptive name like `Helix-Share`?

gd-0: "share" is consistent with bloxroute/ ultrasound so we should keep it the same imo

PR #152, opened 2025-04-29
tracing: iniital support

ltitanb:
add initial support for exporting traces, still experimental and disabled by default. For now only on submit_block and will likely require some adjustments. Next steps would be to progressively trace all endpoints and potentially remove the old traces (currently saved in the db)

### DIFF
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -11,6 +11,7 @@ use alloy_primitives::B256;
 use helix_types::BlsPublicKey;
 use http::HeaderMap;
 use opentelemetry::trace::TracerProvider as _;
+use opentelemetry_otlp::WithExportConfig;
 use reqwest::Url;
 use tracing::error;
 use tracing_appender::{non_blocking::WorkerGuard, rolling::Rotation};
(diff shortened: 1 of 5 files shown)


crates/common/src/utils.rs
```diff
@@ -45,12 +47,31 @@ pub fn init_tracing_log(config: &LoggingConfig) -> WorkerGuard {
...
             let (writer, guard) = tracing_appender::non_blocking(file_appender);
-            let layer = tracing_subscriber::fmt::layer()
+            let file_layer = tracing_subscriber::fmt::layer()
                 .event_format(format)
                 .with_writer(writer)
                 .with_filter(get_crate_filter(log_level));
 
-            tracing_subscriber::registry().with(layer).init();
+            if std::env::var("OTEL_TRACES_EXPORTER") == Ok("true".into()) {
+                println!("starting OTEL exporter");
```
gd-0: is there really a point to this print statement?

crates/common/src/builder_info.rs
```diff
@@ -16,4 +16,8 @@ impl BuilderInfo {
     pub fn can_process_regional_slot_optimistically(&self) -> bool {
         self.is_optimistic && self.is_optimistic_for_regional_filtering
     }
+
+    pub fn builder_id(&self) -> &str {
+        self.builder_id.as_deref().unwrap_or("")
```
gd-0, line 21: Is it not cleaner to return the Option<&str> here then debug print in the tracing trace?

PR #156, opened 2025-05-19
store v3 tx count in header submissions table

vladimir-ea:


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -1376,13 +1376,13 @@ impl DatabaseService for PostgresDatabaseService {
...
+                COALESCE(block_submission.num_txs, header_submission.tx_count) num_txs
             FROM
                 block_submission
             INNER JOIN
-                delivered_payload
-            ON
-                block_submission.block_number = delivered_payload.block_number and block_submission.block_hash = delivered_payload.block_hash
+                delivered_payload ON block_submission.block_number = delivered_payload.block_number and block_submission.block_hash = delivered_payload.block_hash
+            LEFT JOIN
+                header_submission ON block_submission.block_hash = header_submission.block_hash
```
0w3n-d: don't actually need to change this query as, per our discord chat just now, we will always have the block submission num txs here

Owen66 (approved):

PR #157, opened 2025-05-19
limit inflight gossiped header processing

0w3n-d:


crates/api/src/gossip.rs
```diff
@@ -12,9 +14,12 @@ use crate::{
     Api,
 };
 
+const MAX_TASKS_PER_BUILDER: usize = 100;
+
 pub async fn process_gossip_messages<A: Api>(
     builder_api: Arc<BuilderApi<A>>,
     proposer_api: Arc<ProposerApi<A>>,
+    builder_semaphores: Arc<DashMap<BlsPublicKey, Arc<Semaphore>>>,
```
gd-0: why do we need a shared map here? isn't this the only fn that owns it?
```suggestion
    builder_semaphores: HashMap<BlsPublicKey, Arc<Semaphore>>,
```

0w3n-d: fixed now

PR #158, opened 2025-05-20
feat: inclusion lists

jayrabjohns:


crates/api/src/proposer/register.rs
```diff
@@ -70,6 +70,7 @@ impl<A: Api> ProposerApi<A> {
             header_delay: proposer_api.validator_preferences.header_delay,
             delay_ms: proposer_api.validator_preferences.delay_ms,
             gossip_blobs: proposer_api.validator_preferences.gossip_blobs,
+            allow_inclusion_lists: proposer_api.validator_preferences.allow_inclusion_lists,
```
gd-0: allow is the default so would change this to `disable_inclusion_lists`

crates/common/src/api/builder_api.rs
```diff
@@ -43,3 +44,9 @@ pub struct TopBidUpdate {
     pub fee_recipient: Address,
     pub value: U256,
 }
+
+#[derive(Deserialize, Serialize)]
+pub struct InclusionList {
+    #[serde(flatten)]
+    pub bytes: Bytes,
```
gd-0, line 65: should this not be Vec<Bytes> ?

should call the field `txs` or si,milar as well

crates/api/src/builder/inclusion_list.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub async fn get_inclusion_list(
+        Extension(api): Extension<Arc<BuilderApi<A>>>,
+    ) -> impl IntoResponse {
+        if let Some(inclusion_list) = api.curr_slot_info.inclusion_list() {
+            (StatusCode::OK, inclusion_list).into_response()
+        } else {
+            StatusCode::NOT_FOUND.into_response()
+        }
+    }
+}
```
gd-0: this needs to take the slot coordinates (similar to get_header) so we only return the IL if the req is for the relays current head

crates/common/src/config.rs
```diff
@@ -452,3 +458,15 @@ fn test_config() {
     };
     println!("{}", serde_yaml::to_string(&config).unwrap());
 }
+
+#[derive(Clone, Deserialize, Serialize)]
+pub struct InclusionListConfig {
+    pub node: String,
+    pub max_size_bytes: usize,
+}
```
gd-0: can we deser straight into Url?

also we should add the gas limit size of the IL + the weights for time/ price

crates/housekeeper/src/inclusion_list_fetcher.rs
gd-0, line 1: only a single housekeeper runs per cluster. so i think the flow should be. 

- as part of the housekeeper you sub to ILs
- when you receive the IL you write to redis
- get IL api just fetches from redis

crates/housekeeper/src/chain_event_updater.rs
```diff
@@ -241,4 +259,53 @@ impl<D: DatabaseService, A: Auctioneer> ChainEventUpdater<D, A> {
...
+        };
+
+        let Some(inclusion_list) = inclusion_list_opt else {
+            return;
+        };
+
+        let slot: i32 = self.head_slot.try_into().unwrap();
+        let slot_coordinate = get_slot_coordinate(slot, pub_key, parent_hash);
+        let (postgres_result, redis_result) = tokio::join!(
+            self.database.save_inclusion_list(&inclusion_list, slot),
```
0w3n-d: Might be a good not to join on the db task here. We've seen some spikes when writing to the db lately and wouldn't want to delay the completion of this process_slot task.

0w3n-d (approved):

crates/housekeeper/src/chain_event_updater.rs
jayrabjohns: Ah, that's a fair point - we could move the error handling, etc., into the future and leave it as a tokio::spawn, so we don't await it?

PR #162, opened 2025-06-03
feat: change API for inclusion list node to not include metadata

jayrabjohns:
Changing the API we used to fetch inclusion lists to strip out the extra metadata we had there before. Now we just decode the RLP bytes (cheap), extract the metadata we want to store as debug info in postgres, & send on to the simulator for validation.

### DIFF
--- a/crates/common/src/api/builder_api.rs
+++ b/crates/common/src/api/builder_api.rs
@@ -70,7 +70,7 @@ impl TryFrom<InclusionList> for InclusionListWithMetadata {
         for encoded_tx in inclusion_list.txs {
             let decoded_tx = TxEnvelope::decode(&mut encoded_tx.iter().as_slice())?;
             let tx_with_md = InclusionListTxWithMetadata {
-                hash: decoded_tx.hash().clone(),
+                hash: *decoded_tx.hash(),
                 nonce: decoded_tx.nonce(),
                 sender: decoded_tx.recover_signer()?,
                 gas_priority_fee: decoded_tx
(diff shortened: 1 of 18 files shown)


crates/database/src/postgres/migrations/V32__inclusion_lists.sql
```diff
@@ -7,9 +7,8 @@ ADD COLUMN "disable_inclusion_lists" boolean NOT NULL DEFAULT false;
 CREATE TABLE "inclusion_list_txs" (
```
jayrabjohns, line 7: These could be done as a separate migration, but since this isn't deployed on mainnet yet it might just be worth lumping it in here

crates/common/src/api/builder_api.rs
```diff
@@ -45,23 +47,92 @@ pub struct TopBidUpdate {
...
-    pub const fn empty() -> Self {
-        Self { txs: vec![] }
+impl TryFrom<InclusionList> for InclusionListWithMetadata {
+    type Error = InclusionListError;
+
+    fn try_from(inclusion_list: InclusionList) -> Result<Self, Self::Error> {
+        let mut txs = Vec::with_capacity(inclusion_list.txs.len());
+
+        for encoded_tx in inclusion_list.txs {
+            let decoded_tx = TxEnvelope::decode(&mut encoded_tx.iter().as_slice())?;
```
gd-0: i don't think you need the iter() here?

jayrabjohns: ah right, Bytes::as_slice exists but it's private, we could use the [..] slice syntax here instead

gd-0: `&mut other_b.as_ref();`

PR #164, opened 2025-06-03
fix: move inclusion list logic to housekeeper

jayrabjohns:
The logic for fetching & persisting inclusion lists should only be done once across the relay cluster, so it should be moved into the housekeeper.

### DIFF
--- a/crates/housekeeper/src/housekeeper.rs
+++ b/crates/housekeeper/src/housekeeper.rs
@@ -332,8 +332,6 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
             );
         }
 
-        debug!(%epoch, "updating proposer duties");
-
         if validator_registrations.is_empty() {
             warn!(%epoch, "no validator registrationts found");
         } else if let Err(err) =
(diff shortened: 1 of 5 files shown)


0w3n-d (approved):

crates/housekeeper/src/housekeeper.rs
```diff
@@ -524,6 +569,77 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
...
+                return;
+            }
+        };
+
+        let slot_coordinate = get_slot_coordinate(slot, pub_key, parent_hash);
+
+        let (postgres_result, redis_result) = tokio::join!(
+            self.db.save_inclusion_list(&inclusion_list, slot),
+            self.auctioneer.update_current_inclusion_list(inclusion_list.clone(), slot_coordinate)
+        );
```
gd-0, line 605: I'm nbot sure we should block the whole houskeep task when trying to write to the db. Note sure if we have a reasonable timeout on the db write, but if we don't it could block for a while

jayrabjohns: This whole function should be run in a new task anyways - so no worries about blocking the rest of the housekeeper

https://github.com/gattaca-com/helix/pull/164/files/8097423cc08e2623f99e1e65b7077b0c8a9cf602#diff-a6ba36ad27b48a4aed4c3494c22bcb1dc53f15f9ccc8c418bdc8c5820132d5eeR325-R336

PR #166, opened 2025-06-05
chore: cleanup inclusion lists

jayrabjohns:
Move inclusion list fetching & persisting logic into its own module and generally tidy some things up. 

The goal is to encapsulate these things & to make it easier to extend with relay-to-relay communication down the line.


crates/housekeeper/src/inclusion_list/service.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub fn check_eligibility(
+        &self,
+        block_hash: Option<B256>,
+        next_duty: Option<ValidatorRegistrationInfo>,
+    ) -> Option<(B256, BlsPublicKey)> {
+        match (block_hash, next_duty) {
+            (_, None) => {
+                info!("No inclusion list for this slot because we have no registration for the current validator");
+                None
+            }
```
gd-0: we still want to gen an inclusion list here

jayrabjohns: Ah right. We'd still need the current proposer's pubkey and I think the only way to get this is from a proposer registration. Once we have relay-to-relay communication of inclusion lists we could just post that one instead?

gd-0: you still have the proposer pubkey from the original duties. just use that before we filter the duties by registrations

crates/housekeeper/src/housekeeper.rs
```diff
@@ -355,18 +353,20 @@ impl<DB: DatabaseService, A: Auctioneer> Housekeeper<DB, A> {
...
-            file!(),
-            line!(),
-            async move {
-                inclusion_list_service
-                    .handle_inclusion_list_for_slot(block_hash, next_duty, head_slot.as_u64())
-                    .await
-            }
-            .in_current_span(),
-        );
+        if let Some(next_duty) = proposer_duties.iter().find(|duty| duty.slot == head_slot) {
```
gd-0: these duties are post filtering no?

jayrabjohns: pretty sure its the full list, this is actually the list we use to do the filtering https://github.com/gattaca-com/helix/blob/jr/cleanup_inclusion_lists/crates/housekeeper/src/housekeeper.rs#L570

PR #167, opened 2025-06-09
shutdown grace period

vladimir-ea:


crates/common/src/config.rs
```diff
@@ -276,6 +276,10 @@ pub struct StartConfig {
 pub struct RouterConfig {
     #[serde(default)]
     pub enabled_routes: Vec<RouteInfo>,
+    /// On receipt of a shutdown signal, milliseconds to wait after shutting down health endpoint
+    /// and before terminating.
+    #[serde(default)]
+    pub shutdown_delay_ms: u64,
```
Owen66, line 281: Should we set some default here like 30 seconds or something?

Owen66 (approved):

crates/common/src/config.rs
vladimir-ea: or should we default to what we will use ourselves? :-) like maybe 12 seconds?

0w3n-d: Yeah 12 seconds sounds good

PR #178, opened 2025-06-17
adds high priority db pool

0w3n-d:


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -68,18 +68,21 @@ pub struct PostgresDatabaseService {
     validator_pool_cache: Arc<DashMap<String, String>>,
     region: i16,
     pub pool: Arc<Pool>,
+    pub high_priority_pool: Arc<Pool>,
 }
 
 impl PostgresDatabaseService {
     pub fn new(cfg: &Config, region: i16) -> Result<Self, Box<dyn std::error::Error>> {
         let pool = cfg.create_pool(None, NoTls)?;
+        let high_priority_pool = cfg.create_pool(None, NoTls)?;
```
ltitanb, line 77: iirc this creates a pool with some default value of connection, can we sanity check that the values make sense now that we have two pools with the same default?

0w3n-d: default pool size is 16. So it'll be 32 connections per box with 18 boxes now that SG is gone. I think that's fine. Shouldn't overwhelm the db. Can maybe even increase it if needed.

PR #179, opened 2025-06-18
removes locks on header and block submission

0w3n-d:
currently lock rows to udpdate the first seen
if we can accept a slightly less accurate first seen then we can avoid the need for a locks on the row.


0w3n-d:
> i think the bulk of the time is spent checking for conflicts and not necessarily the update no? should we not change so that we can have multiple entries for the same blockhash or will that be a separate PR?

Removing the update removes any lock contention so writes to the table can't block each other anymore. I just want to see what impact that on it's own will have before removing the conflict check. The conflict check is just looking up the hash in the index so I'm not sure it would take so long. If it's still an issue after this then I'll tackle it in a follow up PR.

PR #180, opened 2025-06-19
batch block and header submission writes to the db

0w3n-d:


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -64,11 +74,14 @@ struct TrustedProposerParams {
 pub struct PostgresDatabaseService {
     validator_registration_cache: Arc<DashMap<BlsPublicKey, SignedValidatorRegistrationEntry>>,
     pending_validator_registrations: Arc<DashSet<BlsPublicKey>>,
+    pending_block_submissions: Arc<DashMap<Vec<u8>, PendingBlockSubmissionValue>>,
+    pending_header_submissions: Arc<DashMap<Vec<u8>, PendingHeaderSubmissionValue>>,
```
ltitanb: these are never accessed from multiple threads, we could have a sender here and the receiver in each task. The receiver puts them in a local hashmap which it drains periodically or once after it receives blocks for the following slot

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -64,11 +74,14 @@ struct TrustedProposerParams {
...
     known_validators_cache: Arc<DashSet<BlsPublicKey>>,
     validator_pool_cache: Arc<DashMap<String, String>>,
     region: i16,
     pub pool: Arc<Pool>,
     pub high_priority_pool: Arc<Pool>,
+    processed_slots: Arc<DashSet<i32>>,
```
ltitanb: we can keep this as a local variable and update preferences when an insert for a new slot arrives. Currently we're never popping from the set so it will leak

crates/database/src/postgres/postgres_db_service.rs
ltitanb: if we use the hash as key then we can also use `.entry` and make a check on the "received" timestamp to keep only the one received earlier

0w3n-d: Currently these are accessed from multiple threads. The thread that processes inbound axum request writes to the map and the background task reads from the map and writes to the db. If I'm understanding you correctly, it sounds like you think there should be an extra step to avoid sharing the map across threads? Request handler writes to channel, some additional process reads from the channel and write to a map, and then the background task reads from the map and writes to the db? What's the advantage of adding the channel step in between?

ltitanb: Yep meant that it's not read from multiple threads, just multiple writers and a single reader so a mpsc fits nicely. The dashmap will create unneeded contention  across threads by checking for entries for each block

0w3n-d: OK yeah. Makes sense. I'll change it.

crates/database/src/lib.rs
```diff
@@ -24,6 +24,8 @@ pub async fn start_db_service(config: &RelayConfig) -> eyre::Result<Arc<Postgres
     postgres_db.load_known_validators().await;
     //postgres_db.load_validator_registrations().await;
     postgres_db.start_registration_processor().await;
+    postgres_db.start_block_submission_processor().await;
+    postgres_db.start_header_submission_processor().await;
```
ltitanb: should we move all these in a single `start_processor` function for clarity? Dont think we have a clear use case to start only a few

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -36,6 +42,10 @@ use crate::{
     DatabaseService,
 };
 
+type PendingBlockSubmissionValue = (Arc<SignedBidSubmission>, SubmissionTrace, i16, SystemTime);
+type PendingHeaderSubmissionValue =
+    (Arc<SignedHeaderSubmission>, HeaderSubmissionTrace, Option<u32>, SystemTime);
```
ltitanb: can we either add comments on what the tuple components are or create a struct for it?

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -64,6 +74,8 @@ struct TrustedProposerParams {
 pub struct PostgresDatabaseService {
     validator_registration_cache: Arc<DashMap<BlsPublicKey, SignedValidatorRegistrationEntry>>,
     pending_validator_registrations: Arc<DashSet<BlsPublicKey>>,
+    block_submissions_sender: Option<Sender<PendingBlockSubmissionValue>>,
+    header_submissions_sender: Option<Sender<PendingHeaderSubmissionValue>>,
```
ltitanb, line 88: I would rather not make this optional as realistically we always want to process them

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -246,6 +262,67 @@ impl PostgresDatabaseService {
...
+            let mut ticker = tokio::time::interval(Duration::from_secs(5));
+            let mut last_slot_processed = 0;
+            loop {
+                tokio::select! {
+                    Some(item) = block_submissions_receiver.recv() => {
+                        batch.push(item);
+                        // drain to coalesce
+                        while let Ok(item) = block_submissions_receiver.try_recv() {
+                            batch.push(item);
+                            if batch.len() >= 1_000 { break }
```
ltitanb: why do we need this? the `try_recv` is blocking

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -246,6 +262,67 @@ impl PostgresDatabaseService {
...
+                        }
+                    }
+                    _ = ticker.tick() => {
+                        if !batch.is_empty() {
+                            if let Err(e) = svc_clone._flush_block_submissions(std::mem::take(&mut batch), &mut last_slot_processed).await {
```
ltitanb: `mem::take` reallocates a new vector each time, can we not just pass a reference to the vector and clear it after?

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -430,6 +507,352 @@ impl PostgresDatabaseService {
...
+            params.push(&blk.gas_limit);
+            params.push(&blk.gas_used);
+            params.push(&blk.value);
+            params.push(&blk.num_txs);
+            params.push(&blk.timestamp);
+            params.push(&blk.first_seen);
+        }
+
+        // Build and execute INSERT for block_submission
+        let num_cols = 13;
```
ltitanb: there's a 13 above as well, if it's the same constant can we factor it out in a CONST ?

PR #192, opened 2025-07-01
fix for primev failed update bug

0w3n-d:


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -1195,12 +1195,83 @@ impl DatabaseService for PostgresDatabaseService {
...
+            collateral: PostgresNumeric,
+            is_optimistic: bool,
+            is_optimistic_for_regional_filtering: bool,
+            builder_id: Option<String>,
+            builder_ids: Option<Vec<String>>,
+        }
+
+        let mut structured_builders = Vec::with_capacity(builders.len());
         for builder in builders {
-            self.store_builder_info(&builder.pub_key, &builder.builder_info).await?;
```
jayrabjohns, line 1201: This store_builder_info function isn't being used anywhere else, we could probably remove it now if you'd like

jayrabjohns (approved):
Looks good!

PR #194, opened 2025-07-08
Background redis updates

0w3n-d:


crates/datastore/src/redis/redis_cache.rs
```diff
@@ -145,6 +201,228 @@ impl RedisCache {
         Ok(())
     }
 
+    pub async fn start_payload_attributes_listener(&self) -> Result<(), RedisCacheError> {
```
gd-0, line 198: these fns look basically all the same. Think we can easily make a generic start listener fn. that takes the type that we parse to as `T: DeserializeOwned + Send + Sync`
and also a generic sender

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -633,11 +859,18 @@ impl Auctioneer for RedisCache {
     async fn get_last_slot_delivered(&self) -> Result<Option<u64>, AuctioneerError> {
         let mut record = RedisMetricRecord::new("get_last_slot_delivered");
 
-        let last_slot_delivered =
-            self.get(LAST_SLOT_DELIVERED_KEY).await.map_err(AuctioneerError::RedisError)?;
+        let last_slot_delivered = self.last_delivered_slot.load(Ordering::SeqCst);
```
gd-0: 99% sure there is no need for seqcst ordering here. Was there a specific reason why you used it? also used below.
This adds a lot of overhead as it enforces memory ordering of everything.
Relaxed should be fine here.

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -1015,8 +1267,14 @@ impl Auctioneer for RedisCache {
         builder_pub_key: &BlsPublicKey,
     ) -> Result<BuilderInfo, AuctioneerError> {
         let mut record = RedisMetricRecord::new("get_builder_info");
+        let key = format!("{builder_pub_key:?}");
+        if let Some(cached) = self.builder_info_cache.get(&key) {
+            record.record_success();
+            return Ok(cached);
+        }
```
gd-0: is there a way to include this as part of the `get` function? so we will always try and fetch from a cache if there is one. e.g., could pass an Option<Cache<T>>

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -1094,11 +1360,18 @@ impl Auctioneer for RedisCache {
         proposer_pub_key: &BlsPublicKey,
     ) -> Result<bool, AuctioneerError> {
         let mut record = RedisMetricRecord::new("seen_or_insert_block_hash");
-        let key = get_seen_block_hashes_key(slot, parent_hash, proposer_pub_key);
-        let seen = self.seen_or_add(&key, block_hash).await?;
 
+        let key = get_seen_block_hashes_key(slot, parent_hash, proposer_pub_key, block_hash);
```
ltitanb: why this cant just be the block hash? that way we avoid allocating a string for each submission

crates/datastore/src/redis/redis_cache.rs
0w3n-d: OK I've fixed this

crates/datastore/src/redis/redis_cache.rs
0w3n-d: You mean pass the cache into the hget and then call the get on the cache inside the hget? Is not cleaner the way it atm?

crates/datastore/src/redis/redis_cache.rs
0w3n-d: I did try to create a generic start listener but the functions differ in more ways than just the types, subscriber channel, and cache they insert into or sender in the case of the bids. So I ended up having to pass a handler closure to a very thin generic start listener function.

So I ended up going back to the way it is here. The added complexity of generics and closures and so on just felt a bit too complex just to avoid a bit of repetition.

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -145,6 +201,228 @@ impl RedisCache {
...
     }
 
+    pub async fn start_payload_attributes_listener(&self) -> Result<(), RedisCacheError> {
+        let conn = self.pool.get().await?;
+        let mut pubsub = deadpool_redis::Connection::take(conn).into_pubsub();
+        pubsub.subscribe("payload_attributes_channel").await?;
+
+        let mut message_stream = pubsub.on_message();
+
+        while let Some(msg) = message_stream.next().await {
```
ltitanb, line 205: can this stream ever drop at runtime, if so we should make sure to restart the subscription

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -928,27 +1176,19 @@ impl Auctioneer for RedisCache {
...
+        if let Some(cached) = self.top_bid_value_cache.get(&key) {
+            record.record_success();
+            return Ok(Some(cached));
+        }
 
-        let key = get_builder_latest_bid_value_key(slot, parent_hash, proposer_pub_key);
-        let builder_latest_value = self.hget(&key, &format!("{builder_pub_key:?}")).await?;
+        let top_bid_value = self.get(&key).await?;
+        if let Some(val) = top_bid_value {
+            self.top_bid_value_cache.insert(key.clone(), val);
```
ltitanb:
```suggestion
            self.top_bid_value_cache.insert(key, val);
```

crates/datastore/src/lib.rs
```diff
@@ -39,5 +39,77 @@ pub async fn start_auctioneer(
...
+    });
+
+    let auctioneer_clone = auctioneer.clone();
+    tokio::spawn(async move {
+        loop {
+            if let Err(err) = auctioneer_clone.start_header_tx_root_listener().await {
+                error!("Header tx root listener error: {}", err);
+            }
+        }
+    });
```
ltitanb: i dont think we need this, the tx_root is only checked on `submit_block_v2` which is not on the hot path

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -81,6 +97,13 @@ pub struct RedisCache {
     inclusion_list: broadcast::Sender<InclusionListWithKey>,
     payload_attributes: broadcast::Sender<PayloadAttributesEvent>,
     head_event: broadcast::Sender<HeadEventData>,
+    top_bid_value_cache: Cache<String, U256>,
+    floor_bid_value_cache: Cache<String, U256>,
+    seen_block_hashes: Cache<String, ()>,
+    last_delivered_slot: Arc<AtomicU64>,
+    builder_latest_payload_received_at: Cache<String, u64>,
+    header_tx_root_cache: Cache<String, B256>,
+    builder_info_cache: Cache<String, BuilderInfo>,
```
ltitanb: the key can just be `BlsPublicKey` no need to make it a string

crates/datastore/src/redis/redis_cache.rs
gd-0: thinking is that we don't need to manually add this step in each place. will automatically be done for us when we make new fns etc

crates/datastore/src/redis/redis_cache.rs
0w3n-d: Yeah we've always used slot, parent_hash, proposer_pub_key here with the hash but I agree, it not needed. I don't think we can have the same block hash for more than one slot.

Done

crates/datastore/src/redis/redis_cache.rs
0w3n-d: All of these listeners are call like so just in case

    tokio::spawn(async move {
        loop {
            if let Err(err) = auctioneer_clone.start_best_bid_listener().await {
                error!("Bid listener error: {}", err);
                sleep(Duration::from_secs(5)).await;
            }
        }
    });

crates/datastore/src/redis/redis_cache.rs
0w3n-d: done

crates/datastore/src/lib.rs
0w3n-d: OK removed

crates/datastore/src/redis/redis_cache.rs
0w3n-d: OK done

still need to be serialized and deserialized when published over the pub sub stuff but agree that where we can have the type we should

crates/datastore/src/redis/redis_cache.rs
0w3n-d: OK done

I added this function

```
async fn get_with_cache<T: Clone + Send + Sync + serde::de::DeserializeOwned + 'static>(
        &self,
        key: &str,
        cache: &Cache<String, T>,
    ) -> Result<Option<T>, RedisCacheError> {
        if let Some(cached) = cache.get(key) {
            return Ok(Some(cached));
        }

        let value = self.get::<T>(key).await?;
        if let Some(ref val) = value {
            cache.insert(key.to_string(), val.clone());
        }

        Ok(value)
    }
```

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -794,6 +1025,13 @@ impl Auctioneer for RedisCache {
     ) -> Result<Option<u64>, AuctioneerError> {
         let mut record = RedisMetricRecord::new("get_builder_latest_payload_received_at");
 
+        let key =
+            get_latest_bid_by_builder_key(slot, parent_hash, proposer_pub_key, builder_pub_key);
+        if let Some(cached) = self.builder_latest_payload_received_at.get(&key) {
+            record.record_success();
+            return Ok(Some(cached));
+        }
```
gd-0: can we add a hget with cache varient?

0w3n-d: OK had to refactor a lot of it but that's done now

crates/datastore/src/redis/redis_cache.rs
```diff
@@ -102,7 +125,31 @@ impl RedisCache {
...
+        let top_bid_value_cache =
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(10_000).build();
+        let floor_bid_value_cache =
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(10_000).build();
+        let seen_block_hashes =
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(10_000).build();
+        let builder_latest_payload_received_at =
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(10_000).build();
+        let builder_info_cache =
+            Cache::builder().time_to_idle(Duration::from_secs(300)).max_capacity(10_000).build();
```
vladimir-ea, line 137: do we expect builder info to always be fetched or updated every 5 minutes? is there a danger that an entry will get expired as idle incorrectly?

0w3n-d: Yes fetch_builder_info is called inside the submission handler. So it'll be called on every block submission.

PR #196, opened 2025-07-17
update error response for bid for past slots

0w3n-d:


crates/api/src/builder/submit_block.rs
```diff
@@ -90,6 +90,11 @@ impl<A: Api> BuilderApi<A> {
         // Verify the payload is for the current slot
         if payload.slot() != head_slot + 1 {
             debug!(?block_hash, "submission is for wrong slot");
+
+            if payload.slot() < head_slot + 1 {
+                return Err(BuilderApiError::PastSlotSubmission);
+            }
```
gd-0: lets replace SubmissionForWrongSlot as well. and do submission for past or submission for future. lets also include the current_slot and got in the error

0w3n-d: done

crates/api/src/builder/error.rs
```diff
@@ -166,7 +169,8 @@ impl IntoResponse for BuilderApiError {
             BuilderApiError::HyperError(_) |
             BuilderApiError::AxumError(_) |
             BuilderApiError::PayloadTooLarge { .. } |
-            BuilderApiError::SubmissionForWrongSlot { .. } |
+            BuilderApiError::SubmissionForPastSlot { .. } |
+            BuilderApiError::SubmissionForFutureSlot { .. } |
```
gd-0, line 173: does this into response error remove the text from the resp and just respond with the status code??

crates/api/src/builder/submit_block.rs
```diff
@@ -89,8 +89,14 @@ impl<A: Api> BuilderApi<A> {
 
         // Verify the payload is for the current slot
         if payload.slot() != head_slot + 1 {
-            debug!(?block_hash, "submission is for wrong slot");
-            return Err(BuilderApiError::SubmissionForWrongSlot {
+            if payload.slot() < head_slot + 1 {
```
gd-0: no need to nest here with the payload.slot() != head_slot + 1.

lets just do 
x > y {
} else if  x < y {
}

crates/api/src/builder/error.rs
0w3n-d: No I don't think so. We currently do it this way and see this in the relay logs
```
2025-07-17T17:54:39.029309Z ERROR submit_block{id=26390b2c-c29c-469b-a878-d4766e44d5a8 slot=12162572 builder_pubkey=0xa9d0a0f9059972d775a45d8377768ff20234de91a6fbacba5737ff8803807c38021b5863e5084869e695ef1d6c2fdaef block_hash=0x9dbf16be989b90a43b713ab2edbe0614a9327c7a9f95119f1c34b62f9ad910aa}: helix_api::builder::submit_block: error=submission for wrong slot. expected: 12162572, got: 12162571
```

and this in clickhouse as the response the builder received
```
submission for wrong slot. expected: 12161341, got: 12161340
```

crates/api/src/builder/submit_block.rs
0w3n-d: I originally thought the nested approach would be better since, most of the time, the values are equal and only one conditional check would be needed. But after reading more about branch prediction, it turns out that it doesn’t really matter—the result is the same either way.

I also read that using two separate guard clauses actually performs best (just slightly), since it’s easiest for the CPU to predict, and I think the code looks cleanest that way too. So I went with that.

gd-0: ah right i see your line of thinking. 
by 2 seperarte guard classes you mean no `else if`? didn't know that was faster actually

PR #199, opened 2025-07-20
ensure chain updater can not get stuck

0w3n-d:


crates/housekeeper/src/chain_event_updater.rs
```diff
@@ -216,37 +231,52 @@ impl<A: Auctioneer + 'static> ChainEventUpdater<A> {
...
-                );
-                Some(new_duties)
-            }
-            Err(err) => {
-                error!(error = %err, "Failed to get proposer duties from auctioneer");
-                None
-            }
-        };
+        // Instead of blocking for 1 second, just yield once to allow other tasks to progress.
+        tokio::task::yield_now().await;
```
gd-0: i think the 1s here is needed over a yield. this task could be picked up immediately after this

crates/housekeeper/src/chain_event_updater.rs
```diff
@@ -216,37 +231,52 @@ impl<A: Auctioneer + 'static> ChainEventUpdater<A> {
...
+        let new_duties =
+            match timeout(Duration::from_secs(1), self.auctioneer.get_proposer_duties()).await {
+                Ok(Ok(mut new_duties)) => {
+                    info!(
+                        head_slot = slot,
+                        new_duties = new_duties.len(),
+                        "Fetched proposer duties from auctioneer"
+                    );
+
+                    // --- Parallel Primev proposer checks ---
```
gd-0: these AI style comments just seem useless to me. It's clear there is a timeout and primev is parallel

crates/housekeeper/src/chain_event_updater.rs
```diff
@@ -259,20 +289,31 @@ impl<A: Auctioneer + 'static> ChainEventUpdater<A> {
...
+                }
+            };
 
         // Get the next proposer duty for the new slot.
         let next_duty = self.proposer_duties.iter().find(|duty| duty.slot == slot + 1).cloned();
 
         let update = SlotUpdate { slot: slot.into(), new_duties, next_duty };
-        self.curr_slot_info.handle_new_slot(update, &self.chain_info);
+
+        // Run handle_new_slot in a spawned task so it can't block the select loop.
```
gd-0, line 307: another example of a totally redundant comment. just need to look 3 lines below and there is a tokio spawn

crates/housekeeper/src/chain_event_updater.rs
0w3n-d: OK I've put the sleep back in. I suppose it's unlikely that the sleep is causing the problem.

PR #201, opened 2025-07-21
in mem auction

ltitanb:
move to an in-memory auction to avoid the overhead of going through redis all the time. In practice this means that now get_header and block submissions should happen on the same box

Given the PR is relatively large i tried to keep it simple in the changes, if needed will push some optimizations separately

### DIFF
--- a/crates/common/src/config.rs
+++ b/crates/common/src/config.rs
@@ -390,7 +390,7 @@ impl RouterConfig {
                 "relay is serving headers so should have submissions enabled"
             );
             ensure!(
-                routes.contains(&Route::ProposerApi) || routes.contains(&Route::GetTopBid),
+                routes.contains(&Route::BuilderApi) || routes.contains(&Route::GetTopBid),
                 "routes should have get_top_bid enabled"
             );
 
@@ -401,7 +401,7 @@ impl RouterConfig {
                 "relay is receiving blocks so should have get_header enabled"
             );
             ensure!(
-                routes.contains(&Route::ProposerApi) || routes.contains(&Route::GetTopBid),
+                routes.contains(&Route::BuilderApi) || routes.contains(&Route::GetTopBid),
                 "routes should have get_top_bid enabled"
             );
 
@@ -677,7 +677,7 @@ mod tests {
     #[test]
     fn test_validate_bid_sorter_valid_v3_routes() {
         let config =
-            create_router_config(vec![Route::SubmitHeaderV3, Route::GetHeader, Route::ProposerApi]);
+            create_router_config(vec![Route::SubmitHeaderV3, Route::GetHeader, Route::BuilderApi]);
 
         let result = config.validate_bid_sorter();
         assert!(result.is_ok());


bin/helix/src/main.rs
```diff
@@ -86,12 +88,27 @@ async fn run(config: RelayConfig, keypair: BlsKeypair) -> eyre::Result<()> {
     let db = start_db_service(&config).await?;
     let auctioneer = start_auctioneer(&config, &db).await?;
 
+    let (sorter_tx, sorter_rx) = crossbeam_channel::unbounded();
```
gd-0: should bound this to something reasonable

crates/api/src/builder/api.rs
```diff
@@ -46,11 +46,20 @@ pub struct BuilderApi<A: Api> {
...
     pub _validator_preferences: Arc<ValidatorPreferences>,
     pub current_inclusion_list: Arc<RwLock<Option<InclusionListWithKey>>>,
+    /// Send blocks to the bid sorter
+    pub sorter_tx: crossbeam_channel::Sender<BidSorterMessage>,
+    /// Subscriber for TopBid updates, SSZ encoded
+    pub top_bid_tx: tokio::sync::broadcast::Sender<Bytes>,
+    /// Send headers/blocks to be checked for V2 submissions
+    pub v2_checks_tx: tokio::sync::mpsc::UnboundedSender<V2SubMessage>,
+    /// Set in sorter loop
+    pub shared_floor: Arc<RwLock<U256>>,
```
gd-0, line 60: we should SeqLock this at some point. ready-heavy + copy

crates/api/src/builder/api.rs
```diff
@@ -61,24 +70,36 @@ impl<A: Api> BuilderApi<A> {
...
         relay_config: RelayConfig,
         validator_preferences: Arc<ValidatorPreferences>,
         curr_slot_info: CurrentSlotInfo,
+        sorter_tx: crossbeam_channel::Sender<BidSorterMessage>,
+        top_bid_tx: tokio::sync::broadcast::Sender<Bytes>,
+        v2_checks_tx: tokio::sync::mpsc::UnboundedSender<V2SubMessage>,
+        shared_floor: Arc<RwLock<U256>>,
     ) -> Self {
+        let tx_root_cache =
+            Cache::builder().time_to_live(Duration::from_secs(60)).max_capacity(10_000).build();
```
gd-0: this time-to-live cache is the wrong design for caches that last a single slot. 
Assuming the overhead of this is higher than something like dashmap? if so we should use a dashmap and clear each new slot

crates/api/src/builder/submit_block.rs
```diff
@@ -202,65 +195,33 @@ impl<A: Api> BuilderApi<A> {
...
-                            .store_block_submission(payload, trace, optimistic_version as i16)
-                            .await
-                        {
-                            error!(%err, "failed to store block submission")
-                        }
-                    }
-                    .in_current_span(),
-                );
+        // TODO!!: should we move this before the sorter? if this fails we can't serve the
+        // get_payload Save the execution payload
```
gd-0: this is the same behaviour as now. cost to write this blocking is not worth the minimal chance that something goes very wrong.

We can also add a panic/ exit hook to make sure we don't instantly exit and make sure we only stop once the slot auction is over e.g., at 12 + 4s

crates/api/src/builder/submit_header.rs
```diff
@@ -210,26 +211,22 @@ impl<A: Api> BuilderApi<A> {
...
-                payload.slot().as_u64(),
-                payload.parent_hash(),
-                payload.proposer_public_key(),
-                payload.builder_public_key(),
-                payload.value(),
-                is_cancellations_enabled,
-            )
-            .await?;
+        let floor_bid_value = api.get_current_floor();
+        if payload.value() < floor_bid_value {
```
gd-0: should be `<=` no?

crates/api/src/builder/v2_check.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+    /// Handle valid payload Optimistic V2 demotions checks and demotions.
+    ///
+    /// There are two cases where we might demote a builder here.
+    /// 1) They sent a header but we received no accompanying payload.
+    /// 2) The payload was received > 2 seconds after we received the header.
+    pub async fn run(mut self) {
+        // TODO!!: local telemetry
+
+        let mut tick = tokio::time::interval(Duration::from_secs(5));
```
gd-0: seems an unecassarily long time to wait?

crates/api/src/builder/v2_check.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    /// Check all blocks:
+    /// - for blocks with receive times, check that they are not MAX_DELAY_BETWEEN_V2_SUBMISSIONS_NS
+    ///   apart,
+    /// - for blocks with only header, demote if time is older than
+    ///   MAX_DELAY_BETWEEN_V2_SUBMISSIONS_NS
+    /// - otherwise leave
+    fn check_demotions(&mut self) {
+        let now_ns = utcnow_ns();
+        let mut demoted = HashSet::default();
```
gd-0: allocation crime :)

add this to the loop, pass as a var and clear

crates/api/src/builder/v2_check.rs
```diff
@@ -0,0 +1,194 @@
+use std::{collections::HashMap, sync::Arc, time::Duration};
```
gd-0: if we already have rustc as a dep we should use FxHashMap here

crates/api/src/proposer/get_header.rs
```diff
@@ -127,69 +130,75 @@ impl<A: Api> ProposerApi<A> {
...
+                proposer_api.db.clone(),
+                slot,
+                bid_request.parent_hash,
+                bid_request.pubkey.clone(),
+                bid.header().block_hash().0,
+                trace,
+                mev_boost,
+                user_agent.clone(),
+            )
+            .await;
```
gd-0, line 162: wait, why do we call this in the hot-path?

crates/api/src/proposer/get_header.rs
```diff
@@ -127,69 +130,75 @@ impl<A: Api> ProposerApi<A> {
...
+                    proposer_api
+                        .gossiper
+                        .request_payload(RequestPayloadParams {
+                            slot,
+                            proposer_pub_key: proposer_pubkey_clone,
+                            block_hash,
+                        })
+                        .await
+                });
             }
```
gd-0, line 188: also this in hot-path.

gd-0: ah i guess it's the fact that axum forces the fn to return to actually get the response. is there any way around this?

crates/api/src/test_utils.rs
```diff
@@ -105,6 +114,10 @@ pub fn app() -> Router {
 
 pub fn builder_api_app() -> (Router, Arc<BuilderApi<MockApi>>, CurrentSlotInfo) {
     let current_slot_info = CurrentSlotInfo::new();
+    let (sort_tx, _) = crossbeam_channel::unbounded();
+    let (br_tx, _) = broadcast::channel(1);
+    let (v2_tx, _) = mpsc::unbounded_channel();
```
gd-0: we should really avoid unbounded where poss. we've shot ourselves in the foot with it so many times.
also crossbeam unbounded is a linked list

crates/api/src/proposer/get_header.rs
ltitanb: this spawns a task inside

crates/api/src/proposer/get_header.rs
ltitanb: spawning a single task has minimal overhead so likely fine for now. We can do further optimizations in separate PRs, don't want to change too much pre-existing behaviours

crates/api/src/test_utils.rs
ltitanb: this is for testing only but agreed

crates/api/src/proposer/get_header.rs
```diff
@@ -127,69 +130,75 @@ impl<A: Api> ProposerApi<A> {
...
+
+            let start = Instant::now();
+            let signed_bid = resign_builder_bid(bid, &proposer_api.signing_context, fork);
+            debug!("re-signing builder bid took {:?}", start.elapsed()); // TODO!!: add a metric
+
+            if user_agent.is_some() && is_mev_boost_client(&user_agent.unwrap()) {
+                // Request payload in the background
+                task::spawn(file!(), line!(), async move {
+                    proposer_api
+                        .gossiper
```
0w3n-d, line 180: The original purpose of this gossip was back when we'd gossip headers but no payloads. So we could receive a header in FR from NV and serve the header in FR. But when the payload is called in FR we wouldn't have it. Unless NV was told to gossip it.

Anyway, now that header gossiping is gone this can be removed. (nothing to with this PR but just noticed it so thought I'd record it here.)

crates/api/src/builder/v2_check.rs
ltitanb: used foldhash https://github.com/gattaca-com/helix/pull/201/commits/23bdbfeb1a23c65909baeae9fe6ab95916d77baf

crates/api/src/builder/v2_check.rs
ltitanb: done here https://github.com/gattaca-com/helix/pull/201/commits/23bdbfeb1a23c65909baeae9fe6ab95916d77baf

crates/api/src/builder/v2_check.rs
ltitanb: changed to 3 secs https://github.com/gattaca-com/helix/pull/201/commits/23bdbfeb1a23c65909baeae9fe6ab95916d77baf

crates/api/src/builder/api.rs
ltitanb: that or an atomicU64 with 100 wei precision, will leave for separate PR
https://github.com/gattaca-com/helix/blob/lt/in-mem-auction/crates/common/src/bid_sorter.rs#L194

bin/helix/src/main.rs
ltitanb: removed here: https://github.com/gattaca-com/helix/pull/201/commits/ccdc10a5dd89446739c9e11cff0ec0c4c8bd605a

crates/common/src/bid_sorter.rs
```diff
@@ -0,0 +1,415 @@
...
+        } else {
+            Self { bid_cancel: None, bid_non_cancel: Some(bid) }
+        }
+    }
+
+    /// Returns false if the bid didn't trigger an update
+    fn maybe_update(&mut self, bid: Bid, is_cancellable: bool) -> bool {
+        // check that it's higher than non-cancel bid
+        if let Some(curr_non_cancel) = self.bid_non_cancel {
+            if curr_non_cancel.value > bid.value {
```
0w3n-d: `>=`

crates/common/src/bid_sorter.rs
```diff
@@ -0,0 +1,415 @@
...
+                return false;
+            }
+        }
+
+        if is_cancellable {
+            // keep only if it's the latest
+            if let Some(curr_cancel) = self.bid_cancel {
+                if curr_cancel.on_receive_ns > bid.on_receive_ns {
```
0w3n-d: `>=`

crates/common/src/bid_sorter.rs
```diff
@@ -0,0 +1,415 @@
...
+            shared_best_header,
+            shared_floor,
+        }
+    }
+
+    pub fn run(mut self) {
+        info!("starting bid sorter");
+
+        loop {
+            let Ok(msg) = self.sorter_rx.try_recv() else {
```
0w3n-d, line 269: Why are we spinning here? Can't just block on recv?

crates/api/src/builder/submit_block.rs
```diff
@@ -202,65 +195,33 @@ impl<A: Api> BuilderApi<A> {
...
+        trace!("sent bid to bid sorter");
 
-                // Save submission to db.
-                task::spawn(
-                    file!(),
-                    line!(),
-                    async move {
-                        if let Err(err) = api
-                            .db
-                            .store_block_submission(payload, trace, optimistic_version as i16)
```
0w3n-d, line 246: We're still going to need to save to the db at some point. So we can serve the bids over the data api.

crates/common/src/bid_sorter.rs
ltitanb: spinning is faster, we're going to keep one core dedicated to this which should be ok

crates/api/src/builder/submit_block.rs
ltitanb: that was in the `check_for_later_submissions` , the store is still done here: https://github.com/gattaca-com/helix/blob/lt/in-mem-auction/crates/api/src/builder/submit_block.rs#L241-L252

crates/common/src/bid_sorter.rs
ltitanb: https://github.com/gattaca-com/helix/pull/201/commits/ff820a8fccea140e9597d1f30397248daee09c0f

crates/common/src/bid_sorter.rs
ltitanb: https://github.com/gattaca-com/helix/pull/201/commits/ff820a8fccea140e9597d1f30397248daee09c0f

crates/api/src/builder/v2_check.rs
gd-0: even 3 secs imo. why not every 500ms? that's already an eternity

crates/api/src/proposer/get_header.rs
gd-0: ye but still doing all this work for seemingly no reason? we have the header once we have returned so might as well do all of this after that

crates/api/src/builder/v2_check.rs
ltitanb: what's the advantage in having it lower? if it's higher than 2s there's more chances you're cleaning up the map and not traverse twice the same blocks

crates/api/src/proposer/get_header.rs
ltitanb: this is effectively what is already doing by spawning the task async no? what do you mean all this work?

0w3n-d (approved):

crates/api/src/builder/submit_block.rs
ltitanb: agreed updated comment in https://github.com/gattaca-com/helix/pull/201/commits/22a5ab9841b865cfba9b4e17458d3e942fc81d2e , will leave for a separate PR

crates/api/src/builder/api.rs
ltitanb: changed here to dashmap https://github.com/gattaca-com/helix/pull/201/commits/09b25e3ccd0a7132923670045761b202245224b5

crates/api/src/builder/api.rs
```diff
@@ -46,11 +46,22 @@ pub struct BuilderApi<A: Api> {
...
+    /// Send blocks to the bid sorter
+    pub sorter_tx: crossbeam_channel::Sender<BidSorterMessage>,
+    /// Subscriber for TopBid updates, SSZ encoded
+    pub top_bid_tx: tokio::sync::broadcast::Sender<Bytes>,
+    /// Send headers/blocks to be checked for V2 submissions
+    pub v2_checks_tx: tokio::sync::mpsc::Sender<V2SubMessage>,
+    /// Set in sorter loop
+    pub shared_floor: Arc<RwLock<U256>>,
+    /// Cache of tx roots for v2 submissions
+    pub tx_root_cache: DashMap<(u64, B256), B256>,
```
gd-0: we should have the slot ID in the value so we don't hash it each time

crates/common/src/config.rs
```diff
@@ -359,8 +360,47 @@ impl RouterConfig {
...
+            );
+        }
+
+        if routes.contains(&Route::SubmitBlockOptimistic) {
+            ensure!(routes.contains(&Route::SubmitHeader), "v2 enabled but missing submit header");
+        }
+
+        let is_get_header_instance =
+            routes.contains(&Route::ProposerApi) || routes.contains(&Route::GetHeader);
+        let is_submission_instance =
```
0w3n-d: need to add BuilderApi here as this validation is run before resolve_condensed_routes

crates/api/src/builder/v2_check.rs
gd-0: just to clear the queue out as fast as possible. 99% will have a payload a few millis after so there won't be too much dup work.

Not strongly opinionated here though so happy to go with whatever you think is best

crates/api/src/proposer/get_header.rs
gd-0: well just spawning the task has overhead, also copying all the data etc.
Originally thought to change this we'd just need to move this call lower in the fn. but realise the axum api wants you to return so it's a pain to change. So fine to ignore this.

crates/common/src/config.rs
```diff
@@ -359,8 +360,47 @@ impl RouterConfig {
...
+            routes.contains(&Route::SubmitBlock) || routes.contains(&Route::SubmitHeader);
+
+        if is_get_header_instance {
+            ensure!(
+                is_submission_instance,
+                "relay is serving headers so should have submissions enabled"
+            );
+            ensure!(routes.contains(&Route::GetTopBid), "routes should have get_top_bid enabled");
```
0w3n-d: GetTopBid probably won't be set at this stage. It'll be BuilderApi.

Would be good to have some unit tests here to check all these constraints

crates/common/src/config.rs
ltitanb: https://github.com/gattaca-com/helix/pull/201/commits/e129f3b5672fac31e6bf3c0abb098871a9c84a5a

crates/common/src/config.rs
ltitanb: https://github.com/gattaca-com/helix/pull/201/commits/e129f3b5672fac31e6bf3c0abb098871a9c84a5a

crates/common/src/config.rs
```diff
@@ -379,23 +379,31 @@ impl RouterConfig {
...
+            routes.contains(&Route::SubmitHeaderV3);
 
         if is_get_header_instance {
             ensure!(
                 is_submission_instance,
                 "relay is serving headers so should have submissions enabled"
             );
-            ensure!(routes.contains(&Route::GetTopBid), "routes should have get_top_bid enabled");
+            ensure!(
+                routes.contains(&Route::ProposerApi) || routes.contains(&Route::GetTopBid),
```
0w3n-d: This should BuilderApi rather than ProposerApi as BuilderApi resolves to GetTopBid

PR #205, opened 2025-07-28
fix blocking send 

ltitanb:


crates/housekeeper/src/chain_event_updater.rs
```diff
@@ -315,7 +315,7 @@ impl<A: Auctioneer + 'static> ChainEventUpdater<A> {
         tokio::spawn(async move {
             curr_slot_info.handle_new_slot(update, &chain_info);
         });
-        let _ = self.sorter_tx.send(BidSorterMessage::Slot(slot));
+        let _ = self.sorter_tx.try_send(BidSorterMessage::Slot(slot));
```
0w3n-d, line 318: Maybe we should log the error here?

ltitanb: this will just be queue is full as the sorter is not started on registration boxes

0w3n-d: Oh right. And it would log it for every slot.

PR #212, opened 2025-08-04
feat: relay block merging

MegaRedHand:


crates/datastore/src/redis/redis_cache.rs
```diff
@@ -1132,6 +1135,39 @@ impl Auctioneer for RedisCache {
...
+        merging_preferences: &BlockMergingPreferences,
+    ) -> Result<(), AuctioneerError> {
+        let mut record = RedisMetricRecord::new("save_block_merging_preferences");
+
+        let key = get_cache_block_merging_preferences_key(slot, proposer_pub_key, block_hash);
+        self.set(&key, merging_preferences, Some(BID_CACHE_EXPIRY_S)).await?;
+
+        record.record_success();
+        Ok(())
+    }
```
gd-0: as discussed, i would remove these and just attatch the merging preferences to the header we send to the bid sorter

crates/types/src/bid_submission.rs
```diff
@@ -55,6 +55,8 @@ pub struct SignedBidSubmissionElectra {
     pub execution_payload: Arc<ExecutionPayloadElectra<MainnetEthSpec>>,
     pub blobs_bundle: Arc<BlobsBundle>,
     pub execution_requests: Arc<ExecutionRequests>,
+    #[serde(default)]
+    pub merging_data: BlockMergingData,
```
gd-0: does this work with ssz serialisation? I think this might invalidate any ssz submissions that don't contain the merging data. we might need an entirely new `SignedBidSubmissionElectraMergeable` variant

crates/api/src/builder/api.rs
```diff
@@ -570,6 +570,42 @@ pub(crate) fn sanity_check_block_submission(
     Ok(())
 }
 
+/// Validates all tx indices in the block merging data are within bounds
+pub(crate) fn verify_block_merging_data(
```
gd-0: not sure this really adds much value. If someone sends an invalid tx index we could just skip adding that tx to the mergable data rather than invalidating the entire block

crates/api/src/builder/simulator/multi_simulator.rs
```diff
@@ -90,4 +84,41 @@ impl<A: Auctioneer + 'static, DB: DatabaseService + 'static> MultiSimulator<A, D
...
+                return simulator.process_merge_request(request).await;
+            }
+
+            // If reached max attempts, return an error
+            attempts += 1;
+            if attempts >= self.simulators.len() {
+                return Err(BlockSimError::NoSimulatorAvailable);
+            }
+        }
+    }
```
gd-0: I don't think it makes sense to try and retry a merge like we would for a block validation request. we are going to be very tight on time so either we:
- just try once
- send to all in parallel, race them and take the first result

crates/api/src/builder/simulator/multi_simulator.rs
```diff
@@ -90,4 +84,41 @@ impl<A: Auctioneer + 'static, DB: DatabaseService + 'static> MultiSimulator<A, D
...
+
+    fn next_index(&self) -> usize {
+        // Load balancing: round-robin selection
+        self.next_index
+            .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |x| {
```
gd-0: no need for SeqCst ordering here, this is just a stand-alone count. Relaxed is totally fine

gd-0: ah just noticed this was not your code and you moved from above. Would still be good to change to Relaxed in this PR :)

crates/api/src/builder/submit_block_v2.rs
```diff
@@ -213,6 +220,21 @@ impl<A: Api> BuilderApi<A> {
...
+        // Send the new block merging data to the bid sorter (if it's not the default one)
+        if !mergeable_bundles.bundles.is_empty() || merging_preferences != Default::default() {
+            if let Err(err) = api.sorter_tx.try_send(BidSorterMessage::new_from_payload_submission(
+                &payload,
+                merging_preferences,
+                mergeable_bundles,
+            )) {
+                error!(%err, "failed to send block merging data to bid sorter");
+            }
+        }
```
gd-0: I worry this might have some unexpected side-effects. 

@ltitanb could you take a look at this?

crates/api/src/proposer/get_header.rs
```diff
@@ -101,10 +105,24 @@ impl<A: Api> ProposerApi<A> {
...
         }
 
+        let mut new_bid = None;
+
         // If timing games are enabled for the proposer then we sleep a fixed amount
         // TODO: remove is trusted proposer once people start using header verification
 
         if duty.entry.preferences.header_delay {
+            // Start block merging in the background
+            // TODO: should we wait a bit before block merging?
```
gd-0: we should add a configurable buffer before we would normally fetch the best header.

e.g., 
- configurable buffer is 250ms
- we calculate `sleep_time_ms` to be 1000ms
- we would first sleep to `1000-250` and trigger merging
- sleep another 250ms
- fetch the best header and the merging result

crates/api/src/proposer/get_header.rs
```diff
@@ -146,65 +171,170 @@ impl<A: Api> ProposerApi<A> {
...
+        };
+        let payload_and_blobs = PayloadAndBlobsRef {
+            execution_payload: (&response.execution_payload).into(),
+            blobs_bundle: &response.blobs_bundle,
+        };
+
+        self.auctioneer
+            .save_execution_payload(slot, proposer_pubkey, &block_hash, payload_and_blobs)
+            .await
+            .ok()?;
```
gd-0: this could potentially take a long time so should probably be spawned in a new task

crates/common/src/bid_sorter.rs
```diff
@@ -551,6 +656,56 @@ impl BidSorter {
...
+                                seen_txs.contains(tx) && !b.dropping_txs.contains(&i)
+                            })
+                        })
+                        .cloned()
+                        .collect();
+                seen_txs.extend(bundles.clone().into_iter().flat_map(|b| b.transactions));
+                MergeableBundles { origin: b.origin, bundles }
+            })
+            .collect()
+    }
```
gd-0: Rather than rebuilding the mergable bundles for each best bid which involves a tonne of copying we can just build the map of mergable bundles as we go.

Would look something like this:

BidSorter has a new field: mergeable_bundles which is a map of `hash(MergeableBundle)` -> value of block it was included in and the builder that submitted it. 

That way we keep a deduped map of all the "best" mergeable bundles.

If we make this map shared we can then fetch it once in the get_header call before we start merging

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub dropping_txs: Vec<usize>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct MergeableBundles {
+    /// Address of the builder that submitted these bundles.
+    pub origin: Address,
+    /// List of mergeable bundles.
+    pub bundles: Vec<MergeableBundle>,
+}
```
gd-0: rather than merging all MergeableOrders into MergeableBundles i think it is best to keep the seperation of transactions and bundles internally in the relay. As we extend the merging functionality we will start to introduce more flags that are different for each order type, so i think it's best to keep them logically separated

crates/datastore/src/redis/redis_cache.rs
MegaRedHand: Changed in [1ef129b](https://github.com/gattaca-com/helix/pull/212/commits/1ef129b71a4f6f4d1b5bced8cc9c073484879eb1) 🫡

crates/types/src/bid_submission.rs
MegaRedHand: No, it doesn't. Good catch! Added a manual implementation of the SSZ trait in [ee5f1ca](https://github.com/gattaca-com/helix/pull/212/commits/ee5f1caf99e758cf763910c8723701cfd02bd3a8) to handle cases where the field wasn't specified.

crates/api/src/builder/api.rs
MegaRedHand: Removed in [18bbf8f](https://github.com/gattaca-com/helix/pull/212/commits/18bbf8f7e77f2626416a9b2db42098a466a44719)

crates/types/src/bid_submission.rs
gd-0: nice! could you add a unit test?

MegaRedHand: There's already one:

https://github.com/gattaca-com/helix/blob/6d6e4d939f2428c7f4a09ee8ecd2ac3ae3870224/crates/types/src/bid_submission.rs#L172

crates/api/src/builder/simulator/multi_simulator.rs
MegaRedHand: I agree, but this part only loops through all simulators, searching for any that are enabled, without retries.

crates/api/src/builder/simulator/multi_simulator.rs
MegaRedHand: Done in [4cd328e](https://github.com/gattaca-com/helix/pull/212/commits/4cd328ef22694c83488ac8de4c054fcd2427a8dc)

crates/api/src/proposer/get_header.rs
MegaRedHand: This task can fail. What should we do when it fails? Ignoring it would lead to the payload not being available for a later `get_payload` call.

crates/api/src/proposer/get_header.rs
MegaRedHand: Added in [b15b7e8](https://github.com/gattaca-com/helix/pull/212/commits/b15b7e840e2bbda78c8538d98ab566dde7c76a28)

crates/types/src/block_merging.rs
MegaRedHand: Sounds good. Added in [a02f97b](https://github.com/gattaca-com/helix/pull/212/commits/a02f97b0c1fce2183d2df11e3b148f5e5675092b)

crates/common/src/bid_sorter.rs
MegaRedHand: A map of `hash(MergeableBundle) -> (value of block, builder)` helps with the de-dup, but it would mean we have to sort everything in `get_header`. Is this OK?
An alternative I can think of is, in addition to your deduplication suggestion, keeping the sorted map of `bid -> MergeableBundle` and sharing that with `get_header` instead.

MegaRedHand: Added your suggestion in [2727916](https://github.com/gattaca-com/helix/pull/212/commits/27279161abe4ae2173336e7a2bccb616fe107f9e).

crates/api/src/builder/simulator/multi_simulator.rs
gd-0: ah right missed that. yep all good then

crates/api/src/proposer/get_header.rs
gd-0: I don't think we have any handling in case this call fails in any other place either. We could maybe add a retry here but if redis is dead (only real reason for this to fail IIRC) then there isn't much we can do

crates/common/src/bid_sorter.rs
gd-0: Given that we fetch this in an async task to the normal get header flow i think that's fine. It's unclear to me why we need to sort though?

crates/api/src/builder/api.rs
```diff
@@ -570,6 +573,80 @@ pub(crate) fn sanity_check_block_submission(
...
+                    debug!(
+                        "Got invalid block merging index {}, with a tx count of {}",
+                        tx.index,
+                        txs.len()
+                    );
+                    return None;
+                };
+
+                Some(MergeableTransaction{
+                    transaction: Bytes::from_owner(raw_tx.to_vec()),
```
gd-0: what type is `raw_tx`? Is it Bytes? if so we should just clone here. Bytes::clone is very lightweight as it just clones a ref. Ideally we can avoid copying all of the tx bytes for all mergable orders for every suybmission

gd-0: if `raw_tx` isn't type `Bytes` would you be able to take a look at how easy it is to change?

crates/api/src/builder/submit_block.rs
```diff
@@ -201,14 +202,24 @@ impl<A: Api> BuilderApi<A> {
...
         trace!(is_optimistic = was_simulated_optimistically, "verified submitted block");
 
+        let merging_data = payload.merging_data();
+
+        let mergeable_orders = get_mergeable_orders(&payload, merging_data);
+
+        let merging_preferences =
+            BlockMergingPreferences { allow_appending: merging_data.allow_appending };
+
+        // TODO: send merging preferences here too
```
gd-0: TODO no longer needed

crates/common/src/bid_sorter.rs
MegaRedHand: I was under the impression that we wanted orders to be included in order of highest to lowest bid. If we don't need that, we can avoid the sorting.

crates/common/src/bid_sorter.rs
```diff
@@ -66,21 +78,68 @@ impl BestGetHeader {
...
+        Some((entry.bid, entry.metadata))
     }
 
     fn reset(&self) {
         *self.0.write() = None
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct BestMergeableOrders(Arc<Mutex<HashMap<MergeableOrder, (U256, Address)>>>);
```
gd-0: RwLock probably makes more sense here given get_header calls just need a ref?

crates/api/src/proposer/get_header.rs
MegaRedHand: Done in [c4a9e1e](https://github.com/gattaca-com/helix/pull/212/commits/c4a9e1e69260b29e63b7d6f4ab28b3192e289921)

crates/common/src/bid_sorter.rs
```diff
@@ -66,21 +78,68 @@ impl BestGetHeader {
...
+        orders.into_iter().fold(vec![], |mut acc, (order, (_, origin))| {
+            let last = acc.last_mut();
+            if last.is_none() || last.as_ref().unwrap().origin != origin {
+                acc.push(MergeableOrders::new(origin, vec![order]));
+            } else {
+                last.unwrap().orders.push(order);
+            };
+            acc
+        })
+    }
```
gd-0: Why do we need to sort here? we are sorting by the total block payment which is not what we will sort by when actually merging in (that will be based on the induvidual orders value). i think we can just collect into a vec and that's enough

crates/common/src/bid_sorter.rs
```diff
@@ -66,21 +78,68 @@ impl BestGetHeader {
...
+        // Merge contiguous orders with the same origin
+        orders.into_iter().fold(vec![], |mut acc, (order, (_, origin))| {
+            let last = acc.last_mut();
+            if last.is_none() || last.as_ref().unwrap().origin != origin {
+                acc.push(MergeableOrders::new(origin, vec![order]));
+            } else {
+                last.unwrap().orders.push(order);
+            };
+            acc
+        })
```
gd-0: not sure i understand why we need to merge orders from the same origin? these are induvidual things that should be treated seperately. the origin is just used for attribution at the end once it has been applied

crates/common/src/bid_sorter.rs
```diff
@@ -66,21 +78,68 @@ impl BestGetHeader {
...
+    }
+
+    pub fn insert_orders(&self, bid_value: U256, mergeable_orders: MergeableOrders) {
+        let mut order_map = self.0.lock();
+        let origin = mergeable_orders.origin;
+
+        mergeable_orders.orders.into_iter().for_each(|o| {
+            order_map
+                .entry(o.clone())
```
gd-0: this clones the entire mergable order even if the entry is there. probably better to just use get_mut() and insert()

crates/common/src/bid_sorter.rs
gd-0: added some more detailed comments on this below

crates/api/src/builder/submit_block.rs
MegaRedHand: Thanks! Removed in [18b0dc1](https://github.com/gattaca-com/helix/pull/212/commits/18b0dc16f31067740fbfb9f7e9ab5cdcca9c00ab)

crates/common/src/bid_sorter.rs
MegaRedHand: Changed in [775a3ce](https://github.com/gattaca-com/helix/pull/212/commits/775a3ce77977d3a48749e8ced6bb2c9ddb6b3557)

crates/common/src/bid_sorter.rs
MegaRedHand: I misunderstood. Changed it in [1cd1184](https://github.com/gattaca-com/helix/pull/212/commits/1cd11842a6845a3ab4de0cce0c3136e31beee30a)

crates/common/src/bid_sorter.rs
MegaRedHand: Removed it in [1cd1184](https://github.com/gattaca-com/helix/pull/212/commits/1cd11842a6845a3ab4de0cce0c3136e31beee30a). Also, flattened the structure a bit in [57ce7d3](https://github.com/gattaca-com/helix/pull/212/commits/57ce7d3d18cd690c9dac6b5cac7331fbd60917e9)

crates/common/src/bid_sorter.rs
MegaRedHand: We take ownership of the value, so the clone wasn't needed. Removed it and addressed some other clippy warnings in [c268f25](https://github.com/gattaca-com/helix/pull/212/commits/c268f258b14c008b57559ce8617296667780d7b8)

crates/api/src/builder/api.rs
MegaRedHand: `raw_tx` is a `&VariableList`, so I don't see an easy way to turn that into `Bytes` without cloning it.
We could change the `VariableList` to `Bytes` inside the `ExecutionPayloadElectra` struct, but since it comes from `lh_types`, we'll first need to implement our own and change all uses of it.

gd-0: right ok sounds like a pretty big job. this is fine to keep as is then.
Could you open a github issue for this?

MegaRedHand: Sure! Here: #214

I also noticed the same can be done for blobs in blob transactions so I added that to the issue

MegaRedHand:
Added support for blob transactions in https://github.com/gattaca-com/helix/pull/212/commits/9d0cad6491269ef392f0908d6d00121932babec4 (single commit) to make it easier to review. cc @gd-0

crates/api/src/builder/api.rs
```diff
@@ -570,6 +574,189 @@ pub(crate) fn sanity_check_block_submission(
     Ok(())
 }
 
+/// Expands the references in [`BlockMergingData`] from the transactions in the
+/// payload of the given submission. If any bundle references a transaction not in
+/// the payload, it will be silently ignored.
+pub fn get_mergeable_orders(
```
ltitanb, line 666: this function is quite heavy so we should not have it on the hot path, see also comments below

crates/api/src/builder/api.rs
```diff
@@ -570,6 +574,189 @@ pub(crate) fn sanity_check_block_submission(
...
+        block_blobs_bundles.commitments.iter().map(|c| c.calculate_versioned_hash()).collect();
+    let txs = execution_payload.transactions();
+    let mergeable_orders = merging_data
+        .merge_orders
+        .iter()
+        .map(|order| order_to_mergeable(order, &txs, &blob_versioned_hashes, &block_blobs_bundles))
+        .flatten()
+        .collect();
+
+    MergeableOrders::new(execution_payload.fee_recipient(), mergeable_orders)
```
ltitanb: the `fee_recipient` is not a perfect indicator for the builder address, i would rather we add it explicitly to `BlockMergingData`

crates/api/src/builder/api.rs
```diff
@@ -570,6 +574,189 @@ pub(crate) fn sanity_check_block_submission(
...
+    old_indices
+        .iter()
+        .enumerate()
+        .filter(|(_, tx_index)| flagged_indices.contains(tx_index))
+        .map(|(i, _)| i)
+        .collect()
+}
+
+fn is_blob_transaction(raw_tx: &[u8]) -> bool {
+    raw_tx.get(0).copied().unwrap_or(0) == TxType::Eip4844
```
ltitanb:
```suggestion
    raw_tx.first().is_some_and(|&b| b == TxType::Eip4844)
```

crates/api/src/builder/api.rs
```diff
@@ -570,6 +574,189 @@ pub(crate) fn sanity_check_block_submission(
...
+
+    MergeableOrders::new(execution_payload.fee_recipient(), mergeable_orders)
+}
+
+fn order_to_mergeable(
+    order: &Order,
+    txs: &Transactions,
+    blob_versioned_hashes: &[B256],
+    block_blobs_bundles: &BlobsBundle,
+) -> Option<MergeableOrder> {
```
ltitanb: i think it's better if here we return a `Result` and we only accept the mergeable orders if all orders are Ok, otherwise it's a sign the request was  malformed and would err on the safe side to avoid unbundling

ltitanb: since we don't care too much about the error you can use `eyre` and `bail!`

crates/types/src/bid_submission.rs
```diff
@@ -48,14 +49,130 @@ impl BidTrace {
...
 pub struct SignedBidSubmissionElectra {
     pub message: BidTrace,
     pub execution_payload: Arc<ExecutionPayloadElectra<MainnetEthSpec>>,
     pub blobs_bundle: Arc<BlobsBundle>,
     pub execution_requests: Arc<ExecutionRequests>,
     pub signature: BlsSignature,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "BlockMergingData::is_default")]
+    pub merging_data: BlockMergingData,
+}
```
ltitanb: this makes the new field leak all over and potentially slow down even the normal deserialization without data. I think a cleaner approach is to introduce a new struct:
```rust
struct SignedBidSubmissionWithMergingData {
    submission: SignedBidSubmission,
    merging_data: BlockMergingData
}
```
the builder would set a header eg `x-mergeable` so we know whether to use `SignedBidSubmission` or `SignedBidSubmissionWithMergingData` to decode the submission

crates/api/src/builder/api.rs
```diff
@@ -570,6 +574,189 @@ pub(crate) fn sanity_check_block_submission(
...
+        Order::Tx(tx) => {
+            let Some(raw_tx) = txs.get(tx.index) else {
+                debug!(
+                    "Got invalid block merging index {}, with a tx count of {}",
+                    tx.index,
+                    txs.len()
+                );
+                return None;
+            };
+            let mut blobs_bundle = None;
```
ltitanb: instead of setting a `blobs_bundle` why not return `TxEip4844WithSidecar` directly?

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,176 @@
...
+use serde::{Deserialize, Serialize};
+use ssz_derive::{Decode, Encode};
+
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode)]
+#[ssz(enum_behaviour = "transparent")]
+#[serde(untagged)]
+pub enum Order {
+    Tx(Transaction),
+    Bundle(Bundle),
+}
```
ltitanb: wonder if we actually need this here? why not just bundles, and transactions are bundles with a single tx. This simplifies a bunch the code paths and is cleaner imo

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub can_revert: bool,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, TestRandom)]
+#[serde(deny_unknown_fields)]
+/// All indices are taken from block.transactions.
+pub struct Bundle {
+    /// Signals txs that are part of the bundle
+    /// and ordering of txs.
+    pub txs: Vec<usize>,
```
ltitanb: this vectors are very small, i wonder if we could use something like https://docs.rs/smallvec/latest/smallvec/

crates/api/src/builder/api.rs
```diff
@@ -570,6 +574,189 @@ pub(crate) fn sanity_check_block_submission(
...
+                    Ok(Bytes::from_owner(raw_tx.to_vec()))
+                })
+                .collect::<Result<_, ()>>();
+
+            let Ok(transactions) = transactions_res else {
+                return None;
+            };
+
+            let reverting_txs = update_flagged_indices(&bundle.txs, &bundle.reverting_txs);
+            let dropping_txs = update_flagged_indices(&bundle.txs, &bundle.dropping_txs);
```
ltitanb: why is this step needed? shouldn't the builder already set the correct indices within the bundle?

crates/api/src/builder/simulator/mod.rs
```diff
@@ -57,3 +58,25 @@ impl BlockSimRequest {
         }
     }
 }
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct BlockMergeRequest {
+    /// The original payload value
+    pub value: U256,
```
ltitanb: nit
```suggestion
    pub original_value: U256,
```

crates/api/src/builder/simulator/optimistic_simulator_tests.rs
```diff
@@ -32,10 +32,12 @@ mod simulator_tests {
...
             MockDatabaseService::new(Arc::new(Default::default()), Arc::new(Default::default()));
-        OptimisticSimulator::new(Arc::new(auctioneer), Arc::new(db), http, SimulatorConfig {
-            url: endpoint.to_string(),
-            namespace: "test".to_string(),
-        })
+        OptimisticSimulator::new(
+            Arc::new(auctioneer),
+            Arc::new(db),
+            http,
+            SimulatorConfig { url: endpoint.to_string(), namespace: "test".to_string() },
```
ltitanb: seeing a bunch of these formatting changes, should go away with `just fmt`

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -23,6 +25,31 @@ pub struct BlockSimRpcResponse {
...
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+pub struct BlockMergeResponse {
+    pub execution_payload: ExecutionPayload,
+    pub execution_requests: ExecutionRequests,
+    pub blobs_bundle: BlobsBundle,
+    /// Total value for the proposer
+    pub value: U256,
```
ltitanb: nit
```suggestion
    pub proposer_value: U256,
```

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -23,6 +25,31 @@ pub struct BlockSimRpcResponse {
...
+            RpcResult::Err { error } => Some(error),
+            _ => None,
+        }
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+pub struct BlockMergeResponse {
+    pub execution_payload: ExecutionPayload,
+    pub execution_requests: ExecutionRequests,
```
ltitanb, line 47: is this field actually needed? as in can it change during merging?

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -23,6 +25,31 @@ pub struct BlockSimRpcResponse {
...
+    pub fn as_error(&self) -> Option<&JsonRpcError> {
+        match self {
+            RpcResult::Err { error } => Some(error),
+            _ => None,
+        }
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+pub struct BlockMergeResponse {
```
ltitanb, line 45: if we wanted to optimize this, we already have all the transactions and blobs, so we could only send back the ordering and roots from the simulator. This would save a lot in bandwidth / network

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -192,4 +219,89 @@ impl<DB: DatabaseService + 'static> RpcSimulator<DB> {
...
         }
     }
+
+    pub async fn send_merge_request(
+        &self,
+        request: BlockMergeRequest,
+    ) -> Result<Response, reqwest::Error> {
+        let mut headers = HeaderMap::new();
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
+        headers.insert("X-High-Priority", HeaderValue::from_static("true"));
```
ltitanb: do we need this? assuming we handle all merge requests the same

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -192,4 +219,89 @@ impl<DB: DatabaseService + 'static> RpcSimulator<DB> {
...
+
+        let rpc_payload = json!({
+            "jsonrpc": "2.0",
+            "id": "1",
+            "method": format!("{}_mergeBlockV1", self.simulator_config.namespace),
```
ltitanb: this was only used for simulating blocks for backwards compatibility, this is a new endpoint so can always be `relay`

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -192,4 +219,89 @@ impl<DB: DatabaseService + 'static> RpcSimulator<DB> {
             _ => Ok(false),
         }
     }
+
+    pub async fn send_merge_request(
+        &self,
+        request: BlockMergeRequest,
+    ) -> Result<Response, reqwest::Error> {
+        let mut headers = HeaderMap::new();
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
```
ltitanb: this is not needed since you call `.json(&rpc_payload)` when making the request

crates/api/src/builder/submit_block.rs
```diff
@@ -201,14 +202,23 @@ impl<A: Api> BuilderApi<A> {
         };
         trace!(is_optimistic = was_simulated_optimistically, "verified submitted block");
 
+        let merging_data = payload.merging_data();
+
+        let mergeable_orders = get_mergeable_orders(&payload, merging_data);
```
ltitanb: this function is potentially slow and should be moved off the hot path. We should keep the bid sorter light and move all the merging logic in a separate task that can be turned off if not enabled

crates/api/src/builder/submit_block_v2.rs
ltitanb, line 1: we discussed internally and for now we can remove support in v2 for merging , which should simplify a bit the PR

crates/api/src/proposer/get_header.rs
```diff
@@ -156,11 +162,40 @@ impl<A: Api> ProposerApi<A> {
...
+            );
+            let remaining_sleep_time = if sleep_time > block_merging_buffer {
+                sleep(sleep_time - block_merging_buffer).await;
+                block_merging_buffer
+            } else {
+                sleep_time
+            };
+
+            // Start block merging in the background
+            let block_merging = tokio::spawn({
```
ltitanb: the downside of this is that `get_header` may be called by others other than the proposer, and many times per slot, each triggering a merging request. To avoid overloading we should use a background task that continuously produce merged block and get_header can just read off the latest merged block (probably worth adding some sanity checks to check that it's not too stale)

crates/common/src/bid_sorter.rs
```diff
@@ -27,6 +30,15 @@ type BlsPubkey = [u8; 48];
...
     bid: BuilderBid,
+    metadata: BlockMergingPreferences,
+}
+
+#[derive(Debug, Clone)]
+pub struct BidMetadata {
+    /// Preferences related to block merging.
+    pub merging_preferences: BlockMergingPreferences,
+    /// List of mergeable orders, ordered by bid value
+    pub mergeable_orders: HashMap<MergeableOrder, (U256, Address)>,
```
ltitanb: as mentioned above, let's remove this logic from the sorter to a separate task

crates/types/src/block_merging.rs
gd-0: see convo here: https://github.com/gattaca-com/helix/pull/212#discussion_r2271101656
this was actually how it was originally implemented :)

crates/api/src/builder/api.rs
MegaRedHand: Added it in [b75d6bb](https://github.com/gattaca-com/helix/pull/212/commits/b75d6bb00ca6a46600286d22b4afcfb8f4ec5821)

crates/api/src/builder/api.rs
MegaRedHand: These indices are for the whole block, so we turn them into indices for the bundle. An example `BlockMergingData` would be:

```json
{
  "allow_appending": true,
  "merge_orders": [
    {
      "Tx": { "index": 5, "can_revert": true }
    },
    {
      "Bundle": {
        "txs":           [7, 8, 9],
        "reverting_txs": [7],
        "dropping_txs":  [8]
      }
    }
  ]
}
```

We can change them to be indices for the bundle, of course. That should simplify this part.

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Done in [acedf11](https://github.com/gattaca-com/helix/pull/212/commits/acedf11d1eb4dfd6bfac99e154963bacf033744c)

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Yes. One of the appended transactions could, for example, [initiate a validator exit](https://eips.ethereum.org/EIPS/eip-7002), [which would result in an execution request being appended](https://eips.ethereum.org/EIPS/eip-7685).

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: I'm not sure. The [relay-sim-reth](https://github.com/gattaca-com/relay-sim-reth) simulator doesn't seem to use this header.

Removed in [92d34e4](https://github.com/gattaca-com/helix/pull/212/commits/92d34e41c5126da1b16f0e2bc0b719f7f5b7ecc9).

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Good to know. Changed it in [d03ff9c](https://github.com/gattaca-com/helix/pull/212/commits/d03ff9c383c291f261be3292a1a8746044e139e6)

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Good catch! Removed it from here and also from validation and syncing requests. [61cda45](https://github.com/gattaca-com/helix/pull/212/commits/61cda45604f30357c3f2b40ce4c609b8211c7e2c)

crates/api/src/builder/simulator/optimistic_simulator_tests.rs
MegaRedHand: Thanks for the suggestion! Done in [cc20956](https://github.com/gattaca-com/helix/pull/212/commits/cc209566ce235f294f70bf75a462e6a04068e829)

crates/api/src/builder/api.rs
ltitanb: yep let's do that, would align more with what the builder already has. I would still sanity check that they are all valid and discard the block if the request is malformed

crates/types/src/block_merging.rs
MegaRedHand: We can do that. I think we can also change the `usize` to a `u32`/`u16`, which should improve space efficiency.

crates/api/src/builder/api.rs
MegaRedHand: You mean, change the transaction into its network form? That's another way to do it, but it would mean decoding each transaction and encoding them again when sending them to the simulator.

crates/types/src/block_merging.rs
ltitanb: sounds good, only thing may not be compatible with `Encode` / `Decode`, if not feel free to leave a TODO and keep it out of the PR for future optimization

crates/api/src/builder/api.rs
ltitanb: the blob bundle is only needed at the end when sealing the final block. While sorting, all txs and bundles are in the same "pool" so the blobs could be "attached" to the tx via the `TxEip4844WithSidecar` , instead of keeping them in a blobs bundle

ltitanb: though the blob itself is not used in the EL when simulating the tx, so best approach may depend on how https://github.com/gattaca-com/helix/pull/212#discussion_r2278805301 is implemented

ltitanb: @gd-0 may also have some insight here

crates/api/src/builder/api.rs
MegaRedHand: Done in [d8cad7d](https://github.com/gattaca-com/helix/pull/212/commits/d8cad7d0c5d7e86dac68b04240940b85c2e6b5d2). I used strings for the error variant and log any errors with a level of warning.

crates/api/src/builder/api.rs
MegaRedHand: I previously had index validations, but removed them in favor of just dropping the bundle with invalid indices (https://github.com/gattaca-com/helix/pull/212#discussion_r2270981773). Should I add some checks that indices don't reference txs outside the bundle or block? Also, should I drop the whole block, or just the merging data?

cc @gd-0

MegaRedHand: Added a check for wrong indices in drop/can_revert flags, which should drop the whole merging data in case of a malformed request. [ce9b814](https://github.com/gattaca-com/helix/pull/212/commits/ce9b81411efd45dd26e4f198ba05ce3d271a462c)

crates/common/src/bid_sorter.rs
MegaRedHand: This struct was unused. Removed it in [502780f](https://github.com/gattaca-com/helix/pull/212/commits/502780fbfcb179e3858ccbad59d41710001055f8)

crates/api/src/builder/simulator/mod.rs
MegaRedHand: Done in [8873314](https://github.com/gattaca-com/helix/pull/212/commits/8873314515d7bf4a927134e4953f327e1c22e4cf)

crates/api/src/builder/api.rs
MegaRedHand: We now call it after sending the `BidSorter` message and before storing the payload in the auctioneer.

crates/api/src/builder/submit_block.rs
MegaRedHand: I moved the merging logic to another task, as you suggested. I left the merging preferences inside the bid sorter, since it's just a flag, and moving that would require duplicating the mapping. Let me know if I should move it too.

crates/types/src/block_merging.rs
MegaRedHand: Changed it in [688bee6](https://github.com/gattaca-com/helix/pull/212/commits/688bee6c19898ceb10e4ec8eaec87ca27dd20dec) and [5641cf8](https://github.com/gattaca-com/helix/pull/212/commits/5641cf88dbe3ecebd42ee292c4e08af4411a83fe). It seems the `ssz` crate has an implementation for `SmallVec`s.

For now, I added it with an inline size of 2 elements (which keeps the same size as a `Vec<usize>`), but we could increase this by moving to `u32`/`u16`. I think trying a non-sparse representation like a [bit-set](https://docs.rs/bit-set/latest/bit_set/) might also be useful.

crates/types/src/bid_submission.rs
MegaRedHand: Made the change in [ace0203](https://github.com/gattaca-com/helix/pull/212/commits/ace02031c6ae93f97e8d0695c9765f61464abab7). I had to pass a flag to `decode_payload` since it doesn't have access to the headers.
An alternative could be to accept a "merging" query parameter, same as with cancellations.

crates/api/src/proposer/get_header.rs
MegaRedHand: Just added an initial version of this in [c2cadbc](https://github.com/gattaca-com/helix/pull/212/commits/c2cadbc73cc4934f85894a4d35dbc772bc634e21) that continuously merges txs into the best bid's block. For now, it just sleeps between slots or when no bids are available. We could probably remove those by adding a channel for slot events like the bid sorter has, and by sending new bid events to the process (can be through the same channel). Let me know what you think.

crates/api/src/builder/submit_block_v2.rs
MegaRedHand: Removed changes to this endpoint

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Should I make this change? I can open an issue and leave it for a later PR, otherwise.

crates/api/src/builder/api.rs
ltitanb: disagree with this, we should drop the entire block if the block merging is wrong as it means some logic error on the builder side. We should err on the side of caution

crates/api/src/builder/api.rs
ltitanb: instead of logging each error and returning a static string, it seems cleaner and more idiomatic to return a full result with the validation error. then the first error will exit the function and be logged. Any reason not to do this? either `eyre` or explicit `ValidationError` using `thiserror` could work

crates/api/src/builder/simulator/rpc_simulator.rs
ltitanb: Sounds good let's open an issue and leave for a separate PR

crates/api/src/builder/api.rs
ltitanb: considering that we're sending the full transactions initially, we can avoid the blobs bundle and use the `TxEip4844WithSidecar`, unless there'a good reason not to? @MegaRedHand

crates/api/src/builder/simulator/rpc_simulator.rs
```diff
@@ -192,4 +217,87 @@ impl<DB: DatabaseService + 'static> RpcSimulator<DB> {
             _ => Ok(false),
         }
     }
+
+    pub async fn send_merge_request(
+        &self,
+        request: BlockMergeRequest,
+    ) -> Result<Response, reqwest::Error> {
+        let headers = HeaderMap::new();
```
ltitanb: this is unused now

crates/api/src/builder/submit_block.rs
```diff
@@ -70,8 +72,15 @@ impl<A: Api> BuilderApi<A> {
 
         debug!(%head_slot, timestamp_request_start = trace.receive);
 
+        let has_mergeable_data = headers.get("x-mergeable").is_some();
```
ltitanb: can you extract the header name to a constant?

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,187 @@
...
+        } else {
+            Order::Bundle(Bundle::random_for_test(rng))
+        }
+    }
+}
+
+/// Vector of transaction indices. Aliased for easier use.
+pub type TxIndices = SmallVec<[usize; 2]>;
+
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, TestRandom)]
```
ltitanb:
```suggestion
#[derive(Debug, Clone, Copy, Serialize, Deserialize, Encode, Decode, TestRandom)]
```

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl BlockMergingData {
+    pub fn is_default(&self) -> bool {
+        !self.allow_appending && self.merge_orders.is_empty()
+    }
+}
+
+#[derive(
+    Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TestRandom,
```
ltitanb:
```suggestion
    Debug, Default, PartialEq, Eq, Clone, Copy, Serialize, Deserialize, Encode, Decode, TestRandom,
```

crates/api/src/builder/simulator/rpc_simulator.rs
ltitanb: actually let's still avoid sending the blobs since they are not used for sorting

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Represents a single transaction to be appended into a block atomically.
+/// Note that [`PartialEq`] and [`Hash`] implementations ignore the blobs field.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct MergeableTransaction {
+    /// Transaction that can be merged into the block.
+    pub transaction: Bytes,
+    /// Txs that may revert.
+    pub can_revert: bool,
+    /// Blobs used by the transaction
+    pub blobs_bundle: Option<BlobsBundle>,
```
ltitanb: let's remove this and keep the blobs on the relay only, since the blob is not used while sorting

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,187 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct MergeableBundle {
+    /// List of transactions that can be merged into the block.
+    pub transactions: Vec<Bytes>,
+    /// Txs that may revert.
+    pub reverting_txs: TxIndices,
+    /// Txs that are allowed to be omitted, but not revert.
+    pub dropping_txs: TxIndices,
+    /// Blobs used by the bundle
+    pub blobs_bundle: Option<BlobsBundle>,
```
ltitanb: same as above, let's not send blobs

crates/api/src/builder/api.rs
ltitanb: actually let's just avoid sending the blobs to simreth for merging

crates/common/src/bid_sorter.rs
```diff
@@ -66,14 +67,24 @@ impl BestGetHeader {
         slot: u64,
         parent_hash: &B256,
         _validator_pubkey: &BlsPublicKey,
-    ) -> Option<BuilderBid> {
+    ) -> Option<(BuilderBid, BlockMergingPreferences)> {
```
ltitanb: it doesn't seem we're using this, should we just return the `BuilderBid`?

crates/api/src/proposer/mod.rs
```diff
@@ -81,8 +101,140 @@ impl<A: Api> ProposerApi<A> {
...
+                tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+                continue;
+            };
+            let slot = next_duty.slot.into();
+            let best_header_opt = self.shared_best_header.load_any(slot);
+            if best_header_opt.is_none() || !best_header_opt.as_ref().unwrap().1.allow_appending {
+                tokio::time::sleep(std::time::Duration::from_millis(50)).await;
+                continue;
+            }
+            let (best_bid, _) = best_header_opt.unwrap();
```
ltitanb: nit
```suggestion
            let Some((best_bid, merging_preferences)) = best_header_opt else {
                tokio::time::sleep(std::time::Duration::from_millis(50)).await;
                continue;
            };

            if !merging_preferences.allow_appending {
                tokio::time::sleep(std::time::Duration::from_millis(50)).await;
                continue;
            };
```

crates/common/src/merging_pool.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    pub fn load(&self) -> Vec<MergeableOrderWithOrigin> {
+        let order_map = self.0.read();
+        order_map
+            .iter()
+            .map(|(order, (_, origin))| MergeableOrderWithOrigin::new(*origin, order.clone()))
+            .collect()
+    }
+
+    pub fn insert_orders(&self, bid_value: U256, mergeable_orders: MergeableOrders) {
```
ltitanb: let's add more comments explaining the logic here

crates/common/src/merging_pool.rs
```diff
@@ -0,0 +1,114 @@
...
+        Self::new()
+    }
+}
+
+impl BestMergeableOrders {
+    pub fn new() -> Self {
+        Self(Arc::new(RwLock::new(HashMap::with_capacity(5000))))
+    }
+
+    pub fn load(&self) -> Vec<MergeableOrderWithOrigin> {
```
ltitanb: even though we clear the map on a new slot, i wonder if there can be a race where we end up using orders from the previous slot. To avoid this, we could add the slot to the orders and check that when merging the block

crates/common/src/merging_pool.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl MergingPool {
+    pub fn new(
+        pool_rx: crossbeam_channel::Receiver<MergingPoolMessage>,
+        shared_best_orders: BestMergeableOrders,
+    ) -> Self {
+        Self { pool_rx, shared_best_orders, curr_bid_slot: 0 }
+    }
+
+    pub fn run(mut self) {
```
ltitanb: this loop is not doing much and taking up a full core. How about moving the `process_block_merging` and related functions here? that way we dont need to share the orders across threads but can keep them local

ltitanb: would also add a timestamp to ensure that the base block is not older than X ms (configurable, default 250), otherwise we discard the merged block

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, TestRandom)]
+#[serde(deny_unknown_fields)]
+/// All indices are taken from block.transactions.
+pub struct Bundle {
+    /// Signals txs that are part of the bundle
+    /// and ordering of txs.
+    pub txs: TxIndices,
+    /// Txs that may revert.
+    pub reverting_txs: TxIndices,
```
ltitanb, line 62: let's clarify that these are indices within the bundle itself, and have a `.validate()` function that makes sure they are within bounds (otherwise we should discard the entire merging data)

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,187 @@
...
+        if rng.gen() {
+            Order::Tx(Transaction::random_for_test(rng))
+        } else {
+            Order::Bundle(Bundle::random_for_test(rng))
+        }
+    }
+}
+
+/// Vector of transaction indices. Aliased for easier use.
+pub type TxIndices = SmallVec<[usize; 2]>;
```
ltitanb: just a reasonable max, still very cheap
```suggestion
pub type TxIndices = SmallVec<[usize; 8]>;
```

crates/types/src/bid_submission.rs
ltitanb: flag is perfect

crates/api/src/builder/api.rs
ltitanb: on second thought, if the coinbase is not the builder then the fees of the merged transactions will accrue to an address for which we may have no collateral to pay. So let's also add a validation step that the coinbase is in fact the builder fee recipient. Might be a bit redundant but better to err on the safe side @MegaRedHand

MegaRedHand: OK. Added the check in [84f6233](https://github.com/gattaca-com/helix/pull/212/commits/84f62333d24c2a4f317a16c69e3b21a267273907). In case the builder address and coinbase don't match, we ignore the mergeable orders of that payload.

crates/api/src/builder/api.rs
MegaRedHand: You're right. I went for strings because you mentioned `eyre` and it wasn't an already-added dep, but forgot we already had `thiserror` here. Made the change in [a78f77c](https://github.com/gattaca-com/helix/pull/212/commits/a78f77c0fbbc1e6d9d913e8f488e089a5ef6e197)

crates/api/src/builder/api.rs
MegaRedHand: You mean, we should keep the blobs bundle and tx separated, and avoid sending the blobs to the simulator when merging?

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Opened an issue for the transactions https://github.com/gattaca-com/helix/issues/216. I'll start working on the blobs.

crates/types/src/block_merging.rs
MegaRedHand: Added in [6e59fe6](https://github.com/gattaca-com/helix/pull/212/commits/6e59fe6e3858f53b2bfed211f1924527ae017008)

crates/types/src/block_merging.rs
MegaRedHand: Added in [6e59fe6](https://github.com/gattaca-com/helix/pull/212/commits/6e59fe6e3858f53b2bfed211f1924527ae017008)

crates/api/src/builder/submit_block.rs
MegaRedHand: Sure. Done in [d9b2600](https://github.com/gattaca-com/helix/pull/212/commits/d9b260044290c6c6e08dccd111fda404f62f583d)

crates/api/src/builder/simulator/rpc_simulator.rs
MegaRedHand: Good catch! Removed in [5554347](https://github.com/gattaca-com/helix/pull/212/commits/5554347d11c169b4bca7ad2f3c3f25ce1f44cd69)

crates/common/src/bid_sorter.rs
MegaRedHand: Removed in [c1e0fe7](https://github.com/gattaca-com/helix/pull/212/commits/c1e0fe732518642980754f02239e68404e71a7f9)

crates/common/src/merging_pool.rs
MegaRedHand: Added some comments in [fe01103](https://github.com/gattaca-com/helix/pull/212/commits/fe01103a3b239b3704006a6e9fafc7b50a337f5f).

crates/common/src/merging_pool.rs
MegaRedHand: Added a slot check in [abb8521](https://github.com/gattaca-com/helix/pull/212/commits/abb8521bf34bafaf0bceb0c7cfc4ff715f66350d), similar to the one in the bid sorter.

crates/common/src/merging_pool.rs
MegaRedHand: Merging `process_block_merging` with the merging pool is difficult because one is async and the other is not.

crates/types/src/block_merging.rs
MegaRedHand: Clarified in [94f14f2](https://github.com/gattaca-com/helix/pull/212/commits/94f14f23af6ea91c4a3869b998b15b0249fefd97). I added the validation in [318118a](https://github.com/gattaca-com/helix/pull/212/commits/318118a89c8e11fcd025a505d2874bb77543ee45).

crates/common/src/merging_pool.rs
MegaRedHand: Added the timestamp to discard too stale merged blocks in [4b8ff5b](https://github.com/gattaca-com/helix/pull/212/commits/4b8ff5b04a1803f345054eb61f134653ba7e645e).

MegaRedHand: I also changed the behavior when only a merged bid is available in [a6a1326](https://github.com/gattaca-com/helix/pull/212/commits/a6a13265b990ba9706c117f6149c4d8a89db65d0). Previously, we used the merged bid, but I think discarding it would be preferable, right?

MegaRedHand: About merging `process_block_merging` with the merging pool, I think we could embed the merging pool inside `process_block_merging` by processing new mergeable orders between block merges. This would still require running the merging pool in a new thread, but we can use `tokio::task::spawn_blocking` for that, and the resulting state can be (cheaply) sent between task and thread instead of shared.

ltitanb: why not just change the merging loop to be async instead? instead of spinning and using a full core, and we can use a mpsc channel instead of crossbeam

crates/types/src/block_merging.rs
MegaRedHand: Removed in [25d1e22](https://github.com/gattaca-com/helix/pull/212/commits/25d1e22e7fa9922bc57c1bec5f3ecacd750dddba). We avoid sending blobs to the simulator, and it instead sends us the order and tx index of the blob transactions so that we can append their blobs from the relay.

crates/api/src/builder/api.rs
MegaRedHand: Did that in [25d1e22](https://github.com/gattaca-com/helix/pull/212/commits/25d1e22e7fa9922bc57c1bec5f3ecacd750dddba)

crates/types/src/block_merging.rs
MegaRedHand: Done

crates/common/src/merging_pool.rs
MegaRedHand: Moving the `process_block_merging` outside `ProposerApi` is difficult because it depends on the simulators and some of the methods in `ProposerApi`, so, essentially, it would be duplicating part of it.

For now, I moved the merging pool to `process_block_merging` in [e65384e](https://github.com/gattaca-com/helix/pull/212/commits/e65384ecc90ec91dd4523fa0184b3a25f6ae4377). I still have to change the structures used in the merging pool to avoid copies on `load()`, but let me know what you think of the change.

MegaRedHand: Didn't see the comment above.
We could move to an async channel to make it async, but it won't solve the issue if we keep using `try_recv`, since that's blocking and could stall the runtime. One way we could turn it `async` is to add a `yield_now` call every couple of loops. Is there a reason to avoid the `spawn_blocking` call? It should be cheap since `tokio` keeps a thread pool.

ltitanb: Do we need to use try_recv here though? Why not a tokio select on an async channel instead? Spawn_blocking works well for blocking tasks but I wouldn't use it for long running spin loops

MegaRedHand: We only need `try_recv` if we want to avoid locks and simultaneous access to the mergeable orders in both tasks. I changed to an async channel in [ff72a8a](https://github.com/gattaca-com/helix/pull/212/commits/ff72a8a8c4f4a09536ae7079d07393c7e1dbeb54).

MegaRedHand:
@ltitanb Some metrics I thought might be useful:

On each merged block:
- Number of txs (both total & appended)
- Initial block value
- Value delta
- Time in the slot the block was received

The value (initial + delta) of proposed merged blocks.

Let me know if any of these seem useful, or if you have something else in mind.

crates/types/src/block_merging.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct MergeableOrders {
+    /// Address of the builder that submitted these orders.
+    pub origin: Address,
+    /// List of mergeable orders.
+    pub orders: Vec<MergeableOrder>,
+    /// Blobs used by the orders, keyed by their versioned hash.
+    pub blobs: HashMap<B256, BlobsBundle>,
```
ltitanb: the `BlobsBundle` is a container for multiple blobs so has vectors inside. Here we are keying them by versioned hash so we could simply have a (KzgCommitment, KzgProof, Blob) tuple and avoid the allocation of the vectors

crates/api/src/builder/api.rs
```diff
@@ -52,6 +58,8 @@ pub struct BuilderApi<A: Api> {
     pub current_inclusion_list: Arc<RwLock<Option<InclusionListWithKey>>>,
     /// Send blocks to the bid sorter
     pub sorter_tx: crossbeam_channel::Sender<BidSorterMessage>,
+    /// Send mergeable orders to the merging pool
+    pub pool_tx: tokio::sync::mpsc::Sender<MergingPoolMessage>,
```
ltitanb: nit
```suggestion
    pub merge_pool_tx: tokio::sync::mpsc::Sender<MergingPoolMessage>,
```

crates/api/src/builder/api.rs
```diff
@@ -525,6 +535,155 @@ pub(crate) fn sanity_check_block_submission(
...
+
+    Ok(MergeableOrders::new(merging_data.builder_address, mergeable_orders, blobs))
+}
+
+fn order_to_mergeable(
+    order: Order,
+    txs: &Transactions,
+    blob_versioned_hashes: &[B256],
+    block_blobs_bundles: &BlobsBundle,
+    blobs_in_orders: &mut HashMap<B256, BlobsBundle>,
```
ltitanb: can we simplify this? we extract all blobs from the block (for each we have kzg commitment, proof, blob). We cache all of those sent in the block, and here we only check that they are there vs the blob versioned hash we computed

crates/api/src/builder/simulator/mod.rs
```diff
@@ -57,3 +59,43 @@ impl BlockSimRequest {
...
     }
 }
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct BlockMergeRequestRef<'a> {
+    /// The original payload value
+    pub original_value: U256,
+    pub proposer_fee_recipient: Address,
+    pub execution_payload: ExecutionPayloadRef<'a>,
+    pub blobs_bundle: BlobsBundle,
```
ltitanb: i thought we were not sending blobs anymore?

crates/api/src/proposer/block_merging.rs
```diff
@@ -0,0 +1,469 @@
...
+                    // and insert the metadata
+                    e.insert(OrderMetadata { value: bid_value, index });
+                }
+            }
+        });
+    }
+
+    fn reset(&mut self, slot: u64) {
+        self.current_slot = slot;
+        self.order_map.clear();
```
ltitanb, line 462: should we clear also `best_orders` and `mergeable_blob_bundles`

crates/api/src/proposer/block_merging.rs
```diff
@@ -0,0 +1,469 @@
...
+impl<A: Api> ProposerApi<A> {
+    pub async fn process_block_merging(
+        self: Arc<Self>,
+        mut pool_rx: tokio::sync::mpsc::Receiver<MergingPoolMessage>,
+    ) {
+        let mut best_orders = BestMergeableOrders::new();
+        let mut handle = tokio::spawn({
+            let this = self.clone();
+            async move { this.fetch_base_block().await }
+        });
```
ltitanb:
```suggestion
        let mut handle = self.spawn_base_block_fetch_task();
```

crates/api/src/builder/simulator/mod.rs
MegaRedHand: We aren't receiving them on the simulator side. I made the change, but forgot to remove them from the request. Removed them in [79e1625](https://github.com/gattaca-com/helix/pull/212/commits/79e16259c0e84cc52ad07edc611a639c9b911d96)

crates/types/src/block_merging.rs
MegaRedHand: Done. I used a struct instead of a tuple to make it easier to use.

crates/api/src/builder/api.rs
MegaRedHand: Renamed in [7de381f](https://github.com/gattaca-com/helix/pull/212/commits/7de381f3a575f99d8bea8c1822803998741e26e8)

crates/api/src/builder/api.rs
MegaRedHand: We can. The only downside of doing that would be that we would be storing all blobs, even if they aren't used by any orders.

crates/api/src/proposer/block_merging.rs
MegaRedHand: Good catch. Fixed in [75332df](https://github.com/gattaca-com/helix/pull/212/commits/75332df74fe4edc36d3cd7994ed55c9abdf131d0)

crates/api/src/builder/api.rs
MegaRedHand: Simplified it in [958e234](https://github.com/gattaca-com/helix/pull/212/commits/958e234a557606ad2ea426676c761adf4468a2df)

crates/api/src/builder/api.rs
```diff
@@ -522,6 +533,154 @@ pub(crate) fn sanity_check_block_submission(
...
+            let mergeable_bundle =
+                MergeableBundle { transactions, reverting_txs, dropping_txs }.into();
+            Ok(mergeable_bundle)
+        }
+    }
+}
+
+fn is_blob_transaction(raw_tx: &[u8]) -> bool {
+    raw_tx.first().is_some_and(|&b| b == TxType::Eip4844)
+}
```
gd-0: is the first rlp byte always the type? is there not a rlp header with the len etc at the start?

crates/api/src/proposer/get_header.rs
```diff
@@ -170,65 +170,85 @@ impl<A: Api> ProposerApi<A> {
...
+                .load(bid_request.slot.into(), &bid_request.parent_hash);
+            let max_merged_bid_age_ms =
+                proposer_api.relay_config.block_merging_config.max_merged_bid_age_ms;
+
+            match merged_block_bid {
+                None => bid,
+                // If the current best bid has equal or higher value, we use that
+                Some((_, merged_bid)) if merged_bid.value() <= bid.value() => bid,
+                // If the merged bid is stale, we use the current best bid
+                Some((time, _)) if time < utcnow_ms() - max_merged_bid_age_ms => bid,
```
gd-0: we already just took a timestamp for `trace.best_bid_fetched`, can save a sys call to get the time and just re-use that

crates/api/src/builder/api.rs
MegaRedHand: [First byte is the transaction type or an RLP list header (`>= 0xc0`) in case of legacy transactions](https://eips.ethereum.org/EIPS/eip-2718), so this check works for all valid transactions. Added a comment clarifying this.

crates/api/src/proposer/get_header.rs
MegaRedHand: Done in [817c631](https://github.com/gattaca-com/helix/pull/212/commits/817c631aa34e9d6b82b48a1e5bb056ed3f98de62)

MegaRedHand:
@ltitanb Just synced this, and I can see some format changes in the diff. Running `just fmt` in `develop` makes some changes, so it seems there's some unformatted code there.

crates/api/src/builder/simulator/multi_simulator.rs
```diff
@@ -90,4 +83,41 @@ impl<A: Auctioneer + 'static, DB: DatabaseService + 'static> MultiSimulator<A, D
...
+    pub async fn process_merge_request(
+        &self,
+        request: BlockMergeRequest,
+    ) -> Result<BlockMergeResponse, BlockSimError> {
+        let mut attempts = 0;
+
+        loop {
+            let index = self.next_index();
+
+            if self.is_simulator_enabled(index) {
```
0w3n-d: We're going to have a mix of different simulators in production. Not all of them will support block merging so we'll need to add a flag to the simulator that we can check here.

crates/api/src/proposer/block_merging.rs
```diff
@@ -0,0 +1,476 @@
...
+            value: response.proposer_value,
+            pubkey: mock_public_key_bytes(), // this will be replaced when signing the header
+        };
+
+        // Store the payload in the background
+        let payload_and_blobs = PayloadAndBlobs {
+            execution_payload: response.execution_payload,
+            blobs_bundle: merged_blobs_bundle,
+        };
+        // We just log the errors as we can't really do anything about them
```
0w3n-d: Can remove this comment? No errors being logged here.

crates/common/src/config.rs
```diff
@@ -144,6 +147,16 @@ pub struct TimingGameConfig {
     pub default_client_latency_ms: u64,
 }
 
+#[derive(Serialize, Deserialize, Clone, Default)]
+pub struct BlockMergingConfig {
+    /// Flag to enable this feature.
+    #[serde(default = "default_bool::<true>")]
```
0w3n-d: Think it would be safer to set this to false. We can update it to true in the future when it been fully rolled out.

0w3n-d (approved):
Looks good. Just a couple of comments. But no blockers.
Can be done in a follow up PR.

crates/api/src/builder/simulator/multi_simulator.rs
MegaRedHand: Added in [5a52d8e](https://github.com/gattaca-com/helix/pull/212/commits/5a52d8e0dc0dc2a53472a3d65d0acf1e74cc46a8)

crates/common/src/config.rs
MegaRedHand: Got it. Changed in [14d2127](https://github.com/gattaca-com/helix/pull/212/commits/14d21271e202c7036f5d4e53b617d4af212fd230).

crates/api/src/proposer/block_merging.rs
MegaRedHand: Removed in [924f90f](https://github.com/gattaca-com/helix/pull/212/commits/924f90f765036925a59fb822d39f3d89897e40d9)

PR #213, opened 2025-08-12
add timeout header in get_header call 

ltitanb:
allow proposers to set a custom timeout header during `get_header`. With this we also start using the mev boost request latency header that was unused before. I left the previous behaviour unchanged with the `delay_ms` as that requires to change everything in the db, would do it in a separate PR

### DIFF
--- a/crates/common/src/config.rs
+++ b/crates/common/src/config.rs
@@ -128,27 +128,19 @@ pub struct PostgresConfig {
     pub region_name: String,
 }
 
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Deserialize, Clone, Default)]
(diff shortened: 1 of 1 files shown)


crates/api/src/lib.rs
```diff
@@ -71,3 +71,5 @@ pub trait Api: Clone + Send + Sync + 'static {
     type DatabaseService: DatabaseService;
     type MetadataProvider: MetadataProvider;
 }
+
+pub const HEADER_DELAY_MS: &str = "x-header-delay-ms";
```
gd-0: should this not be the proposer signalling their timeout rather than the delay they want us to add? gives us a lot more freedom and means they don;t need to predict the latency on their side

PR #222, opened 2025-08-28
removing redis

0w3n-d:
Removing redis as it only used on our auction relay so everything can be stored locally.
Housekeeper now only runs on the auction relay (the relay that serves headers)


crates/api/src/proposer/get_payload.rs
```diff
@@ -391,12 +387,8 @@ impl<A: Api> ProposerApi<A> {
         Ok(())
     }
 
-    async fn is_trusted_proposer(
-        &self,
-        public_key: &BlsPublicKey,
-    ) -> Result<bool, ProposerApiError> {
-        let is_trusted_proposer = self.auctioneer.is_trusted_proposer(public_key).await?;
-        Ok(is_trusted_proposer)
+    fn is_trusted_proposer(&self, public_key: &BlsPublicKey) -> bool {
```
gd-0: don't need a fn for this anymore

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+        Ok(())
+    }
+
+    #[instrument(skip_all)]
+    fn save_execution_payload(
+        &self,
+        slot: u64,
+        proposer_pub_key: &BlsPublicKey,
+        block_hash: &B256,
+        execution_payload: PayloadAndBlobsRef,
```
ltitanb: let's avoid passing the ref if we need to do `to_owned` immediately

crates/api/src/builder/submit_block.rs
```diff
@@ -218,21 +212,16 @@ impl<A: Api> BuilderApi<A> {
         // Save the execution payload
         // TODO: if this and similar other calls fail we should stop serving headers and send an
         // alert, not much we can do
```
ltitanb: can remove this

crates/api/src/builder/submit_block.rs
```diff
@@ -218,21 +212,16 @@ impl<A: Api> BuilderApi<A> {
...
+        );
         trace.auctioneer_update = utcnow_ns();
         trace!("saved payload to redis");
 
-        if let Err(err) = api.auctioneer.save_bid_trace(payload.bid_trace()).await {
-            error!(%err, "failed to save bid trace");
-            return Err(BuilderApiError::AuctioneerError(err));
-        }
+        api.auctioneer.save_bid_trace(payload.bid_trace());
         trace!("saved bid trace to redis");
```
ltitanb: here and elsewhere we should remove "redis"

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+        let maybe = {
+            let guard = self.last_delivered_hash.read();
+            *guard
+        };
+
+        maybe
+    }
+}
+
+#[async_trait]
```
ltitanb: can we remove the async trait here?

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+        let last_slot_delivered = self.last_delivered_slot.load(Ordering::Relaxed);
+        if last_slot_delivered == 0 {
+            return None;
+        }
+
+        Some(last_slot_delivered)
+    }
+
+    #[instrument(skip_all)]
+    fn check_and_set_last_slot_and_hash_delivered(
```
ltitanb, line 114: this seems to be used only for tests, if so we can probably remove?

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+    inclusion_list: broadcast::Sender<InclusionListWithKey>,
+
+    seen_block_hashes: Cache<B256, ()>,
+    last_delivered_slot: Arc<AtomicU64>,
+    last_delivered_hash: Arc<RwLock<Option<B256>>>,
+    // TODO: remove dead code. Currently if it's removed the compiler complains
+    _builder_latest_payload_received_at: Cache<String, Arc<DashMap<String, u64>>>,
+    builder_info_cache: DashMap<BlsPublicKey, BuilderInfo>,
+    trusted_proposers: DashMap<BlsPublicKey, ProposerInfo>,
+    execution_payload_cache: Cache<String, PayloadAndBlobs>,
```
ltitanb: we can probably remove all the `String` keys now, since most will anyways be `Hash` (and tuples can be keys too)

ltitanb: i wonder how efficient is the `Cache` with this periodic garbage collection, we could have something custom on a per slot basis depending on needs? can also leave for a follow up PR

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+    types::keys::CURRENT_INCLUSION_LIST_KEY,
+    Auctioneer,
+};
+
+const BID_CACHE_EXPIRY_S: u64 = 45;
+const INCLUSION_LIST_EXPIRY_S: u64 = 45;
+const PAYLOAD_ADDRESS_EXPIRY_S: u64 = 24;
+
+#[derive(Clone)]
+pub struct LocalCache {
```
ltitanb, line 34: should clarify if this is thread safe or not, if so every DashMap and DashSet needs to be `Arc`

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+            Some(cached) => Ok(cached.clone()),
+            None => Err(AuctioneerError::BuilderNotFound { pub_key: builder_pub_key.clone() }),
+        }
+    }
+
+    #[instrument(skip_all)]
+    fn demote_builder(&self, builder_pub_key: &BlsPublicKey) -> Result<(), AuctioneerError> {
+        let _ = self.sorter_tx.try_send(BidSorterMessage::Demotion(builder_pub_key.clone()));
+
+        let mut builder_info = self.get_builder_info(builder_pub_key)?;
```
ltitanb: can we get a mutable reference instead?

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+        }
+        builder_info.is_optimistic = false;
+        builder_info.is_optimistic_for_regional_filtering = false;
+
+        self.builder_info_cache.insert(builder_pub_key.clone(), builder_info.clone());
+        Ok(())
+    }
+
+    #[instrument(skip_all)]
+    fn update_builder_infos(&self, builder_infos: &[BuilderInfoDocument]) {
```
ltitanb, line 232: should we clear the map first? to avoid risking keeping a deleted pubkey in the cache

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+        for builder_info in builder_infos {
+            self.builder_info_cache
+                .insert(builder_info.pub_key.clone(), builder_info.builder_info.clone());
+        }
+    }
+
+    #[instrument(skip_all)]
+    fn seen_or_insert_block_hash(&self, block_hash: &B256) -> bool {
+        if self.seen_block_hashes.contains_key(block_hash) {
```
ltitanb: insert usually returns already whether the element existed or not, and you save one lookup

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+    fn kill_switch_enabled(&self) -> bool {
+        self.kill_switch.load(Ordering::Relaxed)
+    }
+
+    //TODO: Implement kill switch functionality. Need to add an internal api to call this.
+    fn enable_kill_switch(&self) {
+        self.kill_switch.store(true, Ordering::Relaxed);
+    }
+
+    //TODO: Implement kill switch functionality. Need to add an internal api to call this.
```
ltitanb, line 294: how about a table with a row per region on the db?

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+        self.kill_switch.store(false, Ordering::Relaxed);
+    }
+
+    #[instrument(skip_all)]
+    fn update_current_inclusion_list(
+        &self,
+        inclusion_list: InclusionListWithMetadata,
+        slot_coordinate: String,
+    ) -> Result<(), AuctioneerError> {
+        let new_inclusion_list_key = format!("{CURRENT_INCLUSION_LIST_KEY}_{slot_coordinate}");
```
ltitanb: we should always avoid String for keys if there's no need

crates/datastore/src/local/local_cache.rs
gd-0: yeah agreed - think we should just clear this on a per slot basis.

crates/datastore/src/local/local_cache.rs
ltitanb: or a simple endpoint that we curl can also work

crates/datastore/src/local/local_cache.rs
gd-0, line 34: we are still using `String` as the key. Now that we don't use redis we should remove any String keys

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(100_000).build();
+        let _builder_latest_payload_received_at =
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(100_000).build();
+        let builder_info_cache = DashMap::new();
+        let last_delivered_slot = Arc::new(AtomicU64::new(0));
+        let last_delivered_hash = Arc::new(RwLock::new(None));
+        let execution_payload_cache =
+            Cache::builder().time_to_idle(Duration::from_secs(45)).max_capacity(100_000).build();
+
+        let trusted_proposers = DashMap::new();
```
gd-0: we should pre-allocate space for these maps to save early latency hits when it needs to resize. there is a lot of funkiness that needs to go on to resize a dashmap due to the locked buckets etc

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    fn get_last_hash_delivered(&self) -> Option<B256> {
+        let maybe = {
+            let guard = self.last_delivered_hash.read();
+            *guard
+        };
+
+        maybe
+    }
```
gd-0: why not just?
```suggestion
    fn get_last_hash_delivered(&self) -> Option<B256> {
        *self.last_delivered_hash.read()
    }
```

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,643 @@
...
+    ) -> Result<BuilderInfo, AuctioneerError> {
+        match self.builder_info_cache.get(builder_pub_key) {
+            Some(cached) => Ok(cached.clone()),
+            None => Err(AuctioneerError::BuilderNotFound { pub_key: builder_pub_key.clone() }),
+        }
+    }
+
+    #[instrument(skip_all)]
+    fn demote_builder(&self, builder_pub_key: &BlsPublicKey) -> Result<(), AuctioneerError> {
+        let _ = self.sorter_tx.try_send(BidSorterMessage::Demotion(builder_pub_key.clone()));
```
gd-0: this seems pretty critical if it fails? should either blocking send or have a critical error

crates/datastore/src/local/local_cache.rs
gd-0, line 294: are these todos still relevant?

crates/datastore/src/local/local_cache.rs
0w3n-d: It's called here https://github.com/gattaca-com/helix/blob/develop/crates/api/src/proposer/get_payload.rs#L193

crates/datastore/src/local/local_cache.rs
0w3n-d: Yeah maybe this wasn't clear in the todo. By internal api, I meant an endpoint we can curl. But I want it to listen on a port that is not publicly accessible. Like an internal or admin port that we can use for things like this.

0w3n-d: Yeah maybe this wasn't clear in the todo. By internal api, I meant an endpoint we can curl. But I want it to listen on a port that is not publicly accessible. Like an internal or admin port that we can use for things like this.

crates/datastore/src/local/local_cache.rs
0w3n-d: It is thread safe but it does `#[derive(Clone)]` so going to wrap them in `Arc` to avoid to risk of it getting cloned somewhere and the DashMaps being copied.

0w3n-d: It is thread safe but it does `#[derive(Clone)]` so going to wrap them in `Arc` to avoid to risk of it getting cloned somewhere and the DashMaps being copied.

crates/datastore/src/local/local_cache.rs
0w3n-d: Yeah I need to add an endpoint to set the kill_switch as I can't do it direct in redis anymore

crates/datastore/src/local/local_cache.rs
0w3n-d: added error log but will follow up with discord alert maybe in separate PR

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/api/src/proposer/get_payload.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: The primev stuff calls this with an array of only a few updates so I can't just treat this as a full replace.
I can't recall ever deleting a pubkey so don't really need to worry about this causing a mem leak. It's less than 100 pubkeys and unlikely to grow by much. The cash will clear on restart and that will happen frequently enough to prevent a leak too.

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: done in fixup

crates/api/src/builder/submit_block.rs
0w3n-d: done in fixup

crates/api/src/builder/submit_block.rs
0w3n-d: done in fixup

crates/datastore/src/local/local_cache.rs
0w3n-d: This is a bit of a change, given there's no current way to create a PayloadAndBlobs. We can fix it in a follow up PR.

0w3n-d:
> is there anything from the auctioneer that we should make sure to start on the registration boxes?

The only thing that will run on the reg box that interacts with the auctioneer is the chain updater. Had you got something in mind that could go wrong?

crates/datastore/src/local/local_cache.rs
ltitanb: i just meant for the function signature, then the caller if needed can do `PayloadAndBlobsRef::to_owned()` when calling the function

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,603 @@
...
+    fn disable_kill_switch(&self) {
+        self.kill_switch.store(false, Ordering::Relaxed);
+    }
+
+    #[instrument(skip_all)]
+    fn update_current_inclusion_list(
+        &self,
+        inclusion_list: InclusionListWithMetadata,
+        slot_coordinate: SlotCoordinate,
+    ) -> Result<(), AuctioneerError> {
```
ltitanb: we can change this to avoid returning anything

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,603 @@
...
+        *self.proposer_duties.write() = duties;
+    }
+
+    #[instrument(skip_all)]
+    fn get_proposer_duties(&self) -> Vec<BuilderGetValidatorsResponseEntry> {
+        self.proposer_duties.read().clone()
+    }
+
+    #[instrument(skip_all)]
+    fn process_slot(&self, head_slot: u64) {
```
ltitanb, line 327: assume not, but worth double checking that we never risk clearing the cache before it's needed? eg removing an execution payload just before the get_payload is called

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,603 @@
...
+        let (inclusion_list, mut il_recv) = broadcast::channel(1);
+
+        // ensure at least one subscriber is running
+        tokio::spawn(async move { while let Ok(_message) = il_recv.recv().await {} });
+
+        let seen_block_hashes = Arc::new(DashSet::new());
+        let builder_info_cache = Arc::new(DashMap::new());
+        let last_delivered_slot = Arc::new(AtomicU64::new(0));
+        let last_delivered_hash = Arc::new(RwLock::new(None));
+        let execution_payload_cache = Arc::new(DashMap::with_capacity(builder_infos.len()));
```
ltitanb: shouldn't capacity be much higher? since a single builder could send a lot of blocks

0w3n-d: Ah this was so supposed to be for builder_info_cache. Fixed now.

crates/datastore/src/local/local_cache.rs
0w3n-d: To be 100% sure I've moved it down to the end after the current slot info has been updated as that's what is checked in the get payload call

crates/datastore/src/local/local_cache.rs
0w3n-d: done

crates/datastore/src/local/local_cache.rs
```diff
@@ -0,0 +1,600 @@
...
+        let builder_info_cache = Arc::new(DashMap::with_capacity(builder_infos.len()));
+        let last_delivered_slot = Arc::new(AtomicU64::new(0));
+        let last_delivered_hash = Arc::new(RwLock::new(None));
+        let execution_payload_cache = Arc::new(DashMap::new());
+        let trusted_proposers = Arc::new(DashMap::with_capacity(200_000));
+        let payload_address_cache = Arc::new(DashMap::new());
+        let bid_trace_cache = Arc::new(DashMap::new());
+        let primev_proposers = Arc::new(DashSet::new());
+        let kill_switch = Arc::new(AtomicBool::new(false));
+        let proposer_duties = Arc::new(RwLock::new(Vec::new()));
```
gd-0: we should pre-allocate all of these dashmaps/ sets

PR #229, opened 2025-09-02
adds admin api for killswitch

0w3n-d:


crates/api/src/admin_service.rs
```diff
@@ -0,0 +1,100 @@
...
+    Extension(auctioneer): Extension<Arc<A>>,
+) -> Result<impl IntoResponse, StatusCode> {
+    auctioneer.enable_kill_switch();
+    Ok((StatusCode::NO_CONTENT, ()))
+}
+
+async fn disable_kill_switch<A: Auctioneer>(
+    Extension(auctioneer): Extension<Arc<A>>,
+) -> Result<impl IntoResponse, StatusCode> {
+    auctioneer.disable_kill_switch();
```
ltitanb, line 36: worth adding a log here and above

PR #230, opened 2025-09-02
add sequence number header

ltitanb:
add an optional sequence number header for block submission, we will discard a submission if we received a higher sequence for the current slot

### DIFF
--- a/crates/api/src/builder/submit_block.rs
+++ b/crates/api/src/builder/submit_block.rs
@@ -72,13 +72,16 @@ impl<A: Api> BuilderApi<A> {
 
         debug!(%head_slot, timestamp_request_start = trace.receive);
 
-        let skip_sigverify =
-            req.headers().get(HEADER_API_KEY).is_some_and(|key| api.auctioneer.check_api_key(key));
         // Decode the incoming request body into a payload
         let (parts, body) = req.into_parts();
         let (payload, is_cancellations_enabled) =
             decode_payload(&parts.uri, &parts.headers, body, &mut trace).await?;
         ApiMetrics::cancellable_bid(is_cancellations_enabled);
+
+        let skip_sigverify = parts
+            .headers
+            .get(HEADER_API_KEY)
+            .is_some_and(|key| api.auctioneer.validate_api_key(key, payload.builder_public_key()));
 
         let block_hash = payload.message().block_hash;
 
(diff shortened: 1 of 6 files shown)


crates/api/src/builder/api.rs
```diff
@@ -63,6 +64,8 @@ pub struct BuilderApi<A: Api> {
     pub tx_root_cache: DashMap<B256, (u64, B256)>,
     /// Best get header to check the current top bid on simulations
     pub shared_best_header: BestGetHeader,
+    /// Builder pubkey -> (bid_slot, sequence)
```
gd-0: should be a bit more descriptive here. e.g., something like i suggested or even a bit more of an explanation
```suggestion
    /// Builder pubkey -> (bid_slot, largest sequence number)
```

crates/api/src/builder/api.rs
```diff
@@ -83,6 +86,7 @@ impl<A: Api> BuilderApi<A> {
         shared_best_header: BestGetHeader,
     ) -> Self {
         let tx_root_cache = DashMap::with_capacity(1000);
+        let sequence_numbers = DashMap::with_capacity(100);
```
gd-0: why so tight on the capacity here? probably reasonable to assume we get > 100 builder keys. alloc is tiny so can make this 500-1k

crates/api/src/builder/api.rs
```diff
@@ -403,6 +408,53 @@ impl<A: Api> BuilderApi<A> {
...
+    ) -> Result<(), BuilderApiError> {
+        let Some(new_seq) = headers
+            .get(HEADER_SEQUENCE)
+            .and_then(|seq| seq.to_str().ok())
+            .and_then(|seq| seq.parse::<u64>().ok())
+        else {
+            return Ok(())
+        };
+
+        match self.sequence_numbers.entry(builder_pubkey) {
```
gd-0: entry api requires ownership of the value. given BlsPublicKey is 96 bytes it will be faster to do a `contains_mut, else` and pass pubkey by ref. did a micro bench a while ago and the entry api was basically always worse or equal so if the type isn't a fast copy it's usually better to use the ref

crates/api/src/builder/submit_block.rs
```diff
@@ -106,6 +108,12 @@ impl<A: Api> BuilderApi<A> {
             });
         }
 
+        api.check_and_update_sequence_number(
+            payload.builder_public_key().clone(),
+            head_slot + 1,
+            &parts.headers,
+        )?;
```
gd-0: can we not do this before decoding the entire payload? we can do it first thing after reading headers. then if out of sequence we just finish reading the body and exit.

ltitanb: we need the builder pubkey, also just realized that this needs to happen after signature verification

gd-0: ah right good point

PR #234, opened 2025-09-03
add dehydrated submissions

ltitanb:
in dehydrated submission, the builder may replace transactions and blobs with their hash. We keep a local cache of orders seen per slot

### DIFF
--- a/crates/types/src/hydration.rs
+++ b/crates/types/src/hydration.rs
@@ -177,6 +177,12 @@ impl Cache {
     }
 }
 
+impl Default for Cache {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// One cache per builder pubkey
 pub struct HydrationCache {
     caches: FxHashMap<BlsPublicKey, Cache>,
@@ -194,7 +200,7 @@ impl HydrationCache {
     }
 }
 
-impl Default for Cache {
+impl Default for HydrationCache {
     fn default() -> Self {
         Self::new()
     }


crates/api/src/builder/api.rs
```diff
@@ -465,7 +486,33 @@ pub async fn decode_payload(
 
     let decoder = SubmissionDecoder::from_headers(headers);
     let body_bytes = to_bytes(req, MAX_PAYLOAD_LENGTH).await?;
-    let payload = decoder.decode(body_bytes)?;
+
+    let should_hydrate = headers.get(HEADER_HYDRATE).is_some();
+
+    let payload: SignedBidSubmission = if should_hydrate {
```
gd-0: what is the reason for spawning a task to do this? what benefit do we gain over just doing it here?

crates/types/src/hydration.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct BlobItem {
+    proof: KzgProof,
+    commitment: KzgCommitment,
+    blob: Blob,
+}
+
+impl DehydratedBidSubmissionElectra {
+    /// Hydrates the bid submission, adds new orders to the cache. Returns the hydrated bid
+    /// submission and the number of tx cache hits and blob cache hits.
+    pub fn hydrate(
```
gd-0, line 76: we should be careful exiting early here. if there is an error we should still go through and build up our cache with the rest of the payloads we don't have yet

crates/api/src/builder/api.rs
ltitanb: lowers contention by avoid sharing the cache

crates/types/src/hydration.rs
ltitanb: done

crates/api/src/builder/api.rs
gd-0: contention higher than passing everything through a task? Really skeptical that this would be faster than sharing the dashmap

ltitanb: each payload will have say 200-300 checks and each one requires a read or write lock vs a warm cache in the task

gd-0: not 100% sold but happy to go with this :+1:

crates/types/src/hydration.rs
```diff
@@ -0,0 +1,187 @@
...
+    message: BidTrace,
+    execution_payload: ExecutionPayloadElectra,
+    blobs_bundle: DehydratedBlobs,
+    execution_requests: ExecutionRequests,
+    signature: BlsSignature,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode)]
+struct DehydratedBlobs {
+    blobs: Vec<KzgProof>,
```
gd-0: would reword to `proofs`

Issue #217, opened 2025-08-25
Builder bid ssz decoding too expensive


chirag-bgh:
SSZ decoding of builder bids is currently too expensive compared to other implementations . I've benchmarked the SSZ decoding performance using a [deneb payload](https://github.com/flashbots/mev-boost-relay/blob/main/testdata/submitBlockPayloadDeneb_Goerli.ssz.gz) across different implementations:

Here are some results:

- Using [SignedBidSubmissinonDeneb](https://github.com/chirag-bgh/helix/blob/25b16f925a2bc17f1718f01f314c2233db313849/crates/api/benches/decoding_bench.rs#L11-L18) with lighthouse types:

Result: ~3.01 ms per decode
```bash
cargo bench -p helix-api --bench decoding_bench
```

```
SSZ size: 872081 bytes (0.83 MB)
payload_decoding/ethereum_ssz with lh types/deneb_payload
                        time:   [3.0094 ms 3.0130 ms 3.0166 ms]
                        change: [-0.0233% +0.2032% +0.4299%] (p = 0.08 > 0.05)
```

- Using [SignedBidSubmissionDenebAlloy](https://github.com/chirag-bgh/alloy-payload-bench/blob/8bb3493788f8b675818f14e56eb38b5899541644/src/lib.rs#L23-L28) with alloy types:
Result: ~58.5 μs per decode (51x faster!)

```bash
cargo bench 
```
```
SSZ size: 872081 bytes (0.83 MB)
payload_decoding/ethereum_ssz with alloy types/deneb_payload
                        time:   [56.285 µs 58.525 µs 61.322 µs]
                        change: [-6.9690% -3.1644% -0.1986%] (p = 0.06 > 0.05)
```

Also ran similar benchmark for flashbots's mevboost relay, [benchmark function](https://github.com/chirag-bgh/mev-boost-relay/blob/ssz_bench/common/ssz_test.go#L220)
Result: ~113.47 μs per decode
```bash
cd common
go test -bench=BenchmarkDecodingPayload -benchmem
```
```
goos: darwin
goarch: arm64
pkg: github.com/flashbots/mev-boost-relay/common
cpu: Apple M2 Pro
BenchmarkDecodingPayload
    ssz_test.go:222: SSZ size: 872081 bytes (0.83 MB)
BenchmarkDecodingPayload/deneb_ssz
[...]

PR #235, opened 2025-09-04
improve ssz types pt2

ltitanb:
this is pt2 following #218, and a fix for #217 . We now migrated all the "hot path" types to custom ones from lighthouse. There is speedup on decoding, but more importantly it allows us to avoid a lot of cloning
```
execution_payload/custom_ssz_decode
                        time:   [299.33 ns 299.77 ns 300.25 ns]
execution_payload/lighthouse_ssz_decode
                        time:   [1.8156 µs 1.8168 µs 1.8182 µs]
```
The signed bid submission is still slower than alloy due to the decoding/validating of pubkeys and signature in `BidTrace`, when removing that it becomes slightly faster:

```
signed_bid_submission/alloy ssz decode
                        time:   [2.6195 µs 2.6221 µs 2.6245 µs]
signed_bid_submission/custom ssz decode
                        time:   [78.097 µs 78.183 µs 78.280 µs]
signed_bid_submission/custom ssz decode - bls bytes
                        time:   [1.7812 µs 1.7828 µs 1.7847 µs]
```
roughly 30us per pubkey and 15us for the signature in my bench. We can probably optimize there but will leave for a future PR

### DIFF
--- a/crates/types/src/execution_payload.rs
+++ b/crates/types/src/execution_payload.rs
@@ -244,6 +244,7 @@ impl ExecutionPayloadHeader {
 
         Ok(())
     }
+
     pub fn to_lighthouse_electra_header(
         &self,
     ) -> Result<lh_types::ExecutionPayloadHeaderElectra<MainnetEthSpec>, SszError> {


crates/types/src/fields.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// VariableList<u8, E::MaxExtraDataBytes>
+    pub struct ExtraData(Bytes);
+    elem = u8;
+    max  = <MainnetEthSpec as EthSpec>::MaxExtraDataBytes;
+}
+
+impl ExtraData {
+    pub fn to_ssz_type(
+        &self,
+    ) -> Result<VariableList<u8, <MainnetEthSpec as EthSpec>::MaxExtraDataBytes>, SszError> {
```
gd-0, line 36: isn't extra data a fixed size listr? 32 bytes?

ltitanb: nope, it's up to 32 https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#executionpayloadv3

PR #236, opened 2025-09-05
migrate to `BlsPubkeyBytes` / `BlsSignatureBytes`

ltitanb:
use `BlsPubkeyBytes` / `BlsSignatureBytes` in most places. This has a few benefits:
- the bytes version is compressed 48 bytes vs the uncompressed 96, so will be faster to move around and to be used as keys in maps
- deserialization is now faster, should make quite a diff when loading a lot of pubkeys/signatures, eg registrations / known validators

We now uncompress pubkeys and signatures lazily and only if strictly needed to verify a signature

final fix for #217 

```
signed_bid_submission/alloy ssz decode
                        time:   [2.5518 µs 2.5535 µs 2.5553 µs]
signed_bid_submission/custom ssz decode
                        time:   [1.7748 µs 1.7761 µs 1.7777 µs]
```

### DIFF
--- a/crates/api/src/proposer/get_payload.rs
+++ b/crates/api/src/proposer/get_payload.rs
@@ -455,7 +455,7 @@ impl<A: Api> ProposerApi<A> {
     async fn save_delivered_payload_info(
         &self,
         payload: Arc<PayloadAndBlobs>,
-        proposer_public_key: &BlsPublicKeyBytes,
+        proposer_public_key: BlsPublicKeyBytes,
         trace: &GetPayloadTrace,
         user_agent: Option<String>,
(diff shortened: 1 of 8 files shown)


crates/database/src/error.rs
```diff
@@ -49,6 +49,9 @@ pub enum DatabaseError {
     #[error("Crypto error: {0:?}")]
     CryptoError(CryptoError),
 
+    #[error("Invalid bytes")]
+    InvalidBytes,
```
gd-0: not very descriptive?

0w3n-d (approved):

PR #242, opened 2025-09-08
remove optimistic v2

ltitanb:


crates/common/src/bid_submission/mod.rs
```diff
@@ -86,12 +83,12 @@ pub enum OptimisticVersion {
     #[default]
     NotOptimistic = 0,
     V1 = 1,
-    V2 = 2,
+    // V2 = 2,
```
0w3n-d, line 86: Why comment it out rather than remove?

0w3n-d (approved):

crates/common/src/bid_submission/mod.rs
ltitanb: just for future reference

PR #251, opened 2025-09-11
feat: multi-relay inclusion lists

MegaRedHand:
This PR adds a first version of the P2P layer, with support for reaching a rough consensus on inclusion lists between relays.


crates/common/src/config.rs
```diff
@@ -207,6 +209,45 @@ pub struct RelayGossipConfig {
...
+    /// Information on known peers
+    #[serde(default)]
+    pub peers: Vec<P2PPeerConfig>,
+    /// Duration until the first cutoff (t_1)
+    #[serde(default = "default_u64::<2000>")]
+    pub cutoff_1_ms: u64,
+    /// Duration until the second cutoff (t_2)
+    /// Should be higher than [`Self::cutoff_1_ms`]
+    #[serde(default = "default_u64::<4000>")]
+    pub cutoff_2_ms: u64,
```
ltitanb: can you add more info on what the cutoffs mean?

crates/common/src/api/builder_api.rs
```diff
@@ -91,31 +95,33 @@ impl TryFrom<InclusionList> for InclusionListWithMetadata {
     }
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq, Encode, Decode, TreeHash)]
 pub struct InclusionList {
-    pub txs: Vec<Bytes>,
+    pub txs: VariableList<Transaction, <MainnetEthSpec as EthSpec>::MaxTransactionsPerPayload>,
```
ltitanb: You can use `Transactions` from `helix_types`

crates/common/src/config.rs
```diff
@@ -207,6 +209,45 @@ pub struct RelayGossipConfig {
...
 }
 
+#[derive(Default, Serialize, Deserialize, Clone)]
+pub struct P2PConfig {
+    /// Whether P2P functionality is enabled or not
+    #[serde(default = "default_bool::<false>")]
+    pub is_enabled: bool,
+    /// Information on known peers
+    #[serde(default)]
+    pub peers: Vec<P2PPeerConfig>,
```
0w3n-d: Calling this p2p is a bit confusing. It's really only going to be a very small network with handful of statically defined nodes. No peer discover mechanisms or anything else that is commonly found in a p2p system. It's more like a distributed websocket api or mesh net maybe? Or maybe federated relay network?

crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+mod socket;
+
+pub struct P2PApi {
+    broadcast_tx: broadcast::Sender<P2PMessage>,
+    api_requests_tx: mpsc::Sender<P2PApiRequest>,
+    signing_context: Arc<RelaySigningContext>,
+    p2p_config: P2PConfig,
+}
+
+impl P2PApi {
```
0w3n-d: P2PApi here could be something like RelayNetworkService or RelayNetworkManager or something. Api seems out of place in the context something like a p2p or mesh network. Would normally protocol rather than api.

crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        tokio::spawn(this.clone().handle_requests(api_requests_rx));
+        info!(peer_count=%this.p2p_config.peers.len(), "Initialized P2P module");
+        this
+    }
+
+    pub fn is_enabled(&self) -> bool {
+        self.p2p_config.is_enabled && !self.p2p_config.peers.is_empty()
+    }
+
+    pub async fn share_inclusion_list(
```
0w3n-d: This couples the inclusion list stuff directly to the transport. This service or manager should maybe have registered handlers that encapsulate the domain logic like inclusion lists. Leaving the transport to only know about it's message types that wrap to higher level type.

crates/p2p/src/message_handler.rs
```diff
@@ -0,0 +1,213 @@
...
+        RawP2PMessage,
+    },
+    request_handlers::P2PApiRequest,
+    socket::PeerSocket,
+    P2PApi,
+};
+
+struct PeerInfo {
+    pubkey_bytes: BlsPublicKeyBytes,
+    pubkey: BlsPublicKey,
```
0w3n-d: Why is the same data stores here twice in two different formats?

crates/p2p/src/request_handlers.rs
```diff
@@ -0,0 +1,55 @@
...
+use helix_common::api::builder_api::InclusionList;
+use helix_types::BlsPublicKeyBytes;
+use tokio::sync::{mpsc, oneshot};
+
+use crate::{
+    inclusion_lists::service::MultiRelayInclusionListsService, messages::P2PMessage, P2PApi,
+};
+
+#[derive(Debug)]
+pub(crate) enum P2PApiRequest {
```
0w3n-d: Instead of request could we call this a message or event? We don't usually have the concept of a request in a p2p or mesh net. Request kind of feels more like client/server semantics.

crates/p2p/src/request_handlers.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[derive(Debug)]
+pub(crate) struct InclusionListRequest {
+    pub(crate) slot: u64,
+    pub(crate) inclusion_list: InclusionList,
+    pub(crate) result_tx: oneshot::Sender<Option<InclusionList>>,
+}
+
+impl P2PApi {
+    pub(crate) async fn handle_requests(
```
0w3n-d: handle_requests doesn't sound like something that's going to kick off a long run task. Maybe start or run?

crates/p2p/src/request_handlers.rs
```diff
@@ -0,0 +1,55 @@
...
+        self: Arc<Self>,
+        mut api_requests_rx: mpsc::Receiver<P2PApiRequest>,
+    ) {
+        let mut inclusion_lists_service = MultiRelayInclusionListsService::new(self.clone());
```
0w3n-d: Maybe this is fine but I'd be inclined to put something like this in the constructor and call it a handler rather than a service.

crates/p2p/src/request_handlers.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        while let Some(request) = api_requests_rx.recv().await {
+            match request {
+                P2PApiRequest::LocalInclusionList(request) => {
```
0w3n-d: Could wrap these in a IL enum and push all this branching into the MultiRelayInclusionListsService.
As it is now, if we were to add more handlers this would become a very large match.

crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        for peer_config in &this.p2p_config.peers {
+            let peer_pubkey = peer_config.pubkey;
+            // Parse URL and try to turn into a request ahead-of-time, panicking on error
+            let request = url_to_client_request(&peer_config.url);
+            // Verify serialized public key is valid
+            let pubkey = BlsPublicKey::deserialize(peer_pubkey.as_ref())
+                .inspect_err(
+                    |e| error!(err=?e, pubkey=%peer_pubkey, "failed to deserialize peer pubkey"),
+                )
+                .expect("pubkey should be valid");
```
0w3n-d: We generally try to avoid any panics in our code. We should just log an error and continue here.

crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        // If API service drops the channel, return None and log a warning
+        result_rx.await.inspect_err(|_| warn!("response channel was dropped")).ok().flatten()
+    }
+}
+
+fn url_to_client_request(url: &str) -> axum::http::Request<()> {
+    let request = url
+        .into_client_request()
+        .inspect_err(|e| error!(err=?e, %url, "invalid peer URL"))
+        .expect("peer URL in config should be valid");
```
0w3n-d: Again here. We shouldn't panic, just return an error.

crates/p2p/src/message_handler.rs
MegaRedHand: It was to avoid having to convert between both formats on each message. But since we changed signature verification to be done once at connection setup, we don't need the deserialized format anymore. Removed it in [5ef9c24](https://github.com/gattaca-com/helix/pull/251/commits/5ef9c24d0d16bf288a946555c628d1a6afea2753)

crates/p2p/src/lib.rs
MegaRedHand: This is run for the initial setup only. I left some panics in the rest of the setup in order to have it fail fast in case the config is invalid.
I can change those cases to log an error and filter out the invalid peer, if that's better.

crates/common/src/config.rs
MegaRedHand: I think `RelayNetwork` sounds good. I'll rename this

crates/p2p/src/lib.rs
MegaRedHand: I named it like so because it accepts connection requests through the API router in the same way as the other `Api` structs. I agree `RelayNetworkService`/`RelayNetworkManager` could be a good alternative, and we could also split the `P2PApi` plus the actual API endpoint (`p2p_connect`) from the rest of the `RelayNetworkService`, to make this easier to understand. What do you think?

crates/p2p/src/request_handlers.rs
MegaRedHand: Requests are internal, to be sent from the `P2PApi` methods or the `handle_ws_connection` of each peer, to the processing loop in `handle_requests`. I called it "Request" to differentiate it from "Message", which I used specifically for protocol messages exchanged between relays. `Event` sounds like a good alternative, so I'll rename it to that, to keep the distinction between "Request" and "Message"

crates/p2p/src/lib.rs
0w3n-d: Sounds good

crates/p2p/src/lib.rs
0w3n-d: I suppose it's fine the way it is, give is happens during startup. We already panic anyway if the config file is malformed and we can't deserialize into the types defined in the struct. This isn't really any different just slight defered.

crates/p2p/src/request_handlers.rs
MegaRedHand: Renamed it to `run_event_handling_loop`

crates/p2p/src/request_handlers.rs
MegaRedHand: Done in [24df2ae](https://github.com/gattaca-com/helix/pull/251/commits/24df2aef7d5ce5c675f6a7027d77929d5e2aeab8)

crates/p2p/src/lib.rs
MegaRedHand: Decoupling this completely would require dyn traits or generics. I think we should wait until we have a few (at least 2) handlers before generalizing this.

crates/common/src/config.rs
```diff
@@ -214,6 +216,45 @@ pub struct RelayGossipConfig {
...
+        }
+        assert!(self.cutoff_1_ms < self.cutoff_2_ms);
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct RelayNetworkPeerConfig {
+    /// The URL of the peer.
+    /// A valid URL is of the form 'ws://<peer-url>/relay/v1/network'
+    pub url: String,
```
ltitanb: No need for the url to have `relay/v1/network` already, let's validate the config and join the url with the path
```suggestion
    pub url: Url,
```

crates/network/src/inclusion_lists/service.rs
```diff
@@ -0,0 +1,196 @@
...
+    event_handlers::{InclusionListEvent, InclusionListEventInfo, NetworkEvent},
+    inclusion_lists::consensus,
+    messages::{InclusionListMessage, InclusionListMessageInfo, NetworkMessage},
+    RelayNetworkManager,
+};
+
+pub(crate) struct MultiRelayInclusionListsService {
+    network_api: Arc<RelayNetworkManager>,
+    cutoff_1: Duration,
+    cutoff_2: Duration,
```
ltitanb: what are these cutoffs?

crates/network/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+                .inspect_err(
+                    |e| error!(err=?e, pubkey=%peer_pubkey, "failed to deserialize peer pubkey"),
+                )
+                .expect("pubkey should be valid");
+
+            // If the peer's pubkey is less than ours, don't try to connect.
+            // Imposing an order on the pubkeys prevents redundant connections between peers.
+            if peer_pubkey > this.signing_context.pubkey {
+                tokio::spawn(this.clone().connect_to_peer(request, peer_pubkey));
+            }
```
ltitanb, line 67: is the assumption here that all peers will have the full list of peers? otherwise there might be gaps

MegaRedHand: We assume each peer in our config knows us. If they didn't, and we tried to connect, we'd get rejected because they don't recognize our pubkey. If both peers know each other, only one of them will try to connect to the other.

We don't, however, assume all peers know each other. Some of them might know only a subset of the peers, but it's important the subset they know also knows about them.

crates/common/src/config.rs
MegaRedHand: Sure. I added a brief explanation on what the parameter does, along with a more extensive explanation of the consensus algorithm in the network's IL module.

crates/common/src/api/builder_api.rs
MegaRedHand: Thanks. Changed it in [c4d8451](https://github.com/gattaca-com/helix/pull/251/commits/c4d84518979aa0f5ed5ffc33ce8eb24391966602)

crates/common/src/config.rs
MegaRedHand: Changed in [7741887](https://github.com/gattaca-com/helix/pull/251/commits/7741887c00ffbcc3c076f86ba99d7083bbcfe476)

crates/network/src/inclusion_lists/service.rs
MegaRedHand: They are the points in time inside the slot when we compute the next inclusion list of the slot (shared or final, respectively). I added some general documentation on `crates/network/src/inclusion_lists/mod.rs` and also added a brief doc comment to each of the fields in this struct.

PR #253, opened 2025-09-15
refactor simulator

ltitanb:
Currently we just spawn a simulation round robin, regardless of how many sims we're currently doing or if the submission is high enough to be considered as top bid.
in this PR:
- we track pending sims per simulator and we cap the max number of sims
- we only keep one sim per builder pubkey

left a few performance related TODOs, will likely leave for follow up PRs

### DIFF
--- a/crates/api/src/builder/simulator/client.rs
+++ b/crates/api/src/builder/simulator/client.rs
@@ -108,7 +108,7 @@ impl SimulatorClient {
 
         match res {
             RpcResult::Err { error } => Err(BlockSimError::BlockValidationFailed(error.message)),
-            RpcResult::Ok(response) => Ok(response),
+            RpcResult::Ok { result } => Ok(result),
         }
     }
 
--- a/crates/api/src/builder/simulator/mod.rs
+++ b/crates/api/src/builder/simulator/mod.rs
@@ -112,7 +112,7 @@ pub struct BlockSimRpcResponse {
 #[derive(serde::Serialize, serde::Deserialize, Debug)]
 #[serde(untagged)]
 pub enum RpcResult<T> {
-    Ok(T),
+    Ok { result: T },
     Err { error: JsonRpcError },
 }
 


crates/api/src/builder/api.rs
```diff
@@ -79,14 +80,19 @@ pub struct BuilderApi<A: Api> {
     pub sequence_numbers: DashMap<BlsPublicKeyBytes, (Slot, u64)>,
     /// Hydration task sender
     pub hydration_tx: tokio::sync::mpsc::Sender<HydrationMessage>,
+    /// Failsafe: if we fail to demote we pause all optimistic submissions
+    pub failsafe_triggered: Arc<AtomicBool>,
+    /// Failsafe: if we fail don't have any synced client we pause all optimistic submissions
+    pub accept_optimistic: Arc<AtomicBool>,
```
gd-0, line 86: if this is in the hotpath and they are always called together we can put both behind a single atomicu8. then just use bitflags

crates/api/src/builder/api.rs
```diff
@@ -313,7 +335,139 @@ impl<A: Api> BuilderApi<A> {
...
+        {
+            self.failsafe_triggered.store(true, Ordering::Relaxed);
+            error!(
+                builder=%builder_public_key,
+                err=%err,
+                "Failed to demote builder in database"
+            );
+        }
+
+        // TODO: do we even need this?
```
0w3n-d: Totally agree we should just drop this. The same data is stored in the demotion table.
Maybe at one point we didn't have the reason in the demotion table.

Currently the only benefit I suppose is the sim fails that don't cause demotion.

crates/api/src/builder/simulator/manager.rs
```diff
@@ -0,0 +1,317 @@
...
+            "sim manager telemetry"
+        )
+    }
+}
+
+/// Pending requests, we only keep the last one for each builder
+struct PendingRquests {
+    map: Vec<SimulatorRequest>,
+    sort_keys: Vec<(u8, u8, u64)>,
+    builder_pubkeys: Vec<BlsPublicKeyBytes>,
```
gd-0, line 263: I think we should avoid having pubkey "maps" in vecs. builderpubkeys can be generated essentially for free. Yes limited by number of blocks you can send, but i'd assume that's pretty high and not worth the additional nanos to use the hashmap

crates/api/src/builder/simulator/manager.rs
```diff
@@ -0,0 +1,317 @@
...
+            builder_pubkeys: Vec::with_capacity(capacity),
+            sort_keys: Vec::with_capacity(capacity),
+        }
+    }
+
+    fn store(&mut self, req: SimulatorRequest, local_telemetry: &mut LocalTelemetry) {
+        if let Some((i, _)) =
+            self.builder_pubkeys.iter_mut().enumerate().find(|(_, r)| **r == *req.builder_pubkey())
+        {
+            if req.on_receive_ns > self.map[i].on_receive_ns {
```
gd-0, line 279: this assumes cancels are always enabled.  not sure if that pr is in yet?
not the end of the world here but we should make sure that we get the cancel remove PR in with this before we deploy

crates/common/src/config.rs
```diff
@@ -191,6 +191,9 @@ pub struct SimulatorConfig {
     pub namespace: String,
     #[serde(default = "default_bool::<false>")]
     pub is_merging_simulator: bool,
+    /// roughly number of cores on simulator
+    #[serde(default)]
+    pub max_concurrent_tasks: Option<usize>,
```
gd-0: when is this ever none? should we not just force it and have a reasonable serde default?

crates/api/src/service.rs
```diff
@@ -83,14 +60,19 @@ pub async fn run_api_service<A: Api>(
     let validator_preferences = Arc::new(config.validator_preferences.clone());
 
     let (gossip_sender, gossip_receiver) = tokio::sync::mpsc::channel(10_000);
-
     let (merge_pool_tx, pool_rx) = tokio::sync::mpsc::channel(10_000);
 
+    let (sim_requests_tx, sim_requests_rx) = mpsc::channel(10_000);
+    let (merge_requests_tx, merge_requests_rx) = mpsc::channel(10_000);
```
gd-0, line 66: just a side note to this pr.

wonder how much unecassary latency we add by using tokio channels. their impl is somewhat cursed. should at a min swap to kanal which IIRC is a few % better. And ultimately get rid of all channels using async

crates/api/src/builder/simulator/manager.rs
ltitanb: not sure what you mean here? are you saying we should use hashmaps instead? or just not have any data structure which keys by builder pubkey?

gd-0: yep meaning we should use a HashMap rather than a vec that we iter.find as a map equivalant

crates/common/src/config.rs
ltitanb: sure, don't have a feeling what's good default here. @0w3n-d any suggestion?

crates/api/src/service.rs
ltitanb: agreed though the sim requests themselves are not on the hot path so a few 10s of us don't make a difference. That being said i will move to a single core without channels so dont think it's worth to add a new dependency for something that will be removed soon anyways

crates/api/src/builder/api.rs
ltitanb: removed here [0d9dc8f](https://github.com/gattaca-com/helix/pull/253/commits/0d9dc8fcc9727b49ed60d67b240fa1e09a12b00b)

crates/api/src/builder/simulator/manager.rs
ltitanb: over the past 7 days we got a total of 40 unique pubkeys, dont think it will be an issue and iterating hashmaps is not great because it's O(capacity). we have similar vecs in a few other places and dont think it caused any problems for that. We could add a per slot max number of non-optimistic submission keys eg 100 as a check

crates/api/src/builder/api.rs
ltitanb: this feels a bit like a micro optimization and will make the code less readable, will move both of these to `bool` flags in a single thread so think it's ok for now

gd-0: optimistic coding

crates/api/src/builder/simulator/manager.rs
gd-0: point is this is an attack vector rather than a happy case issue

gd-0: if we have 100 items. what is the latency gain for a FxHashMap::get vs iter.any ? probably significantly less than the additional inter core latency of having the additional atomic read :)

PR #255, opened 2025-09-17
remove non-cancellable bids

ltitanb:
the vast majority of bids we receive is cancellable. With the new simulator PR we don't gain much by rejecting bids "below floor" as we already have a priority queue for sims (which can be improved independently). Removing support for non-cancellable bids allows us to simplify the submission flow

### DIFF
--- a/crates/database/src/postgres/postgres_db_service.rs
+++ b/crates/database/src/postgres/postgres_db_service.rs
@@ -621,7 +621,7 @@ impl PostgresDatabaseService {
                 decode: item.trace.decode as i64,
                 pre_checks: item.trace.pre_checks as i64,
                 signature: item.trace.signature as i64,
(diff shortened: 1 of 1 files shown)


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -621,7 +621,7 @@ impl PostgresDatabaseService {
                 decode: item.trace.decode as i64,
                 pre_checks: item.trace.pre_checks as i64,
                 signature: item.trace.signature as i64,
-                floor_bid_checks: item.trace.floor_bid_checks as i64,
+                floor_bid_checks: item.trace.simulation as i64,
```
gd-0: this makes it look like floor_bid_checks has the whole sim latency. should be `item.trace.signature` imo so the latency is 0

Issue #191, opened 2025-06-23
Unknown registrations return 200 instead of 400 error response


0w3n-d:
When attempting to register an unknown entity, the system currently responds with a 200 status code, which is misleading. The node operator may think the registration was successful, but it actually was not. 

**Expected behavior:**
- The API should return a 400 error response when an unknown registration is attempted.

**Actual behavior:**
- The API returns a 200 status code even when the registration is unknown and fails.

**Impact:**
- This may cause node operators to believe a registration went through successfully when it did not, potentially leading to operational confusion and misconfiguration.

**Steps to reproduce:**
1. Attempt to register an unknown entity
2. Observe that the API returns a 200 response

**Proposed solution:**
- Return a 400 error code for unknown registration attempts to accurately reflect the error state.

PR #261, opened 2025-09-21
fix: return 400 for unknown validator registrations (#191)

m-v-kalashnikov:
- Proposer API now returns ProposerApiError::NoValidatorsCouldBeRegistered when a batch yields zero successful registrations, surfacing as HTTP 400.
- Added a unit test harness using the mock database’s new “treat all pubkeys as known” toggle and tidied the proposer test utilities accordingly.

Based on commit from [PR#260](https://github.com/gattaca-com/helix/pull/260)

### DIFF
--- a/crates/api/src/proposer/register.rs
+++ b/crates/api/src/proposer/register.rs
@@ -121,7 +121,7 @@ impl<A: Api> ProposerApi<A> {
         let known_pub_keys = proposer_api.db.check_known_validators(registration_pub_keys).await?;
 
         // Check each registration
-        let mut valid_registrations = Vec::with_capacity(known_pub_keys.len());
+        let mut registrations_to_save = Vec::with_capacity(known_pub_keys.len());
 
         let mut handles = Vec::with_capacity(registrations.len());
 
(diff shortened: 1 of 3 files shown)


crates/api/src/proposer/register.rs
```diff
@@ -171,6 +171,17 @@ impl<A: Api> ProposerApi<A> {
 
         let successful_registrations = valid_registrations.len();
 
+        if successful_registrations == 0 {
```
0w3n-d, line 173: successful_registrations can be zero even if it was a valid request. We do a check to see if the timestamp of the registration is greater than what we currently have, if it's not we skip. This can happen for every registration in the request but that shouldn't error.

0w3n-d (commented):
The title of the PR is return 400 for unknown validator registrations but I don't see and check for unknown_registrations being added

crates/api/src/proposer/register.rs
m-v-kalashnikov: Added an explicit `unknown_registrations` guard in `register_validators`; we now return `NoValidatorsCouldBeRegistered` only when every entry in the batch is unknown. At the same time, a batch that’s already up to date (all entries skipped by `is_registration_update_required`) returns `200 OK`. Covered both paths with `all_unknown_registrations_return_bad_request` and the new `up_to_date_registrations_return_ok` tests.

ltitanb:
Hi, we went through quite a refactor and as a result this is now stale. On top of that we refactored the registrations part which i think should already cover this. Apologies for not having reviewed this sooner

PR #265, opened 2025-09-28
return to single core

ltitanb:
refactor many of the caches / flows to go through a new auctioneer running in a single core. This has several benefits:
- easier to reason about pipelines / state dynamics (when the relay should or shouldn't process a given request)
- clearer split between sync / cpu work / async
- easier to reason about what should be shared / Arc 'd / cloned. A lot of the current ones we do can be avoided
- better performance by avoiding lots of locks and sharing

A few things got removed that in an ideal world would have been in separate PRs:
- remove request_payload in gossip
- remove broadcasters
- some support for v3

Given that the PR is already relatively large I left some TODOs that will be done in follow up PRs:
- block merging is still a separate task but will be moved to the auctioneer
- auctioneer will be moved to its own crate
- components like the bid sorter are "stitched together" to minimize code changes, but a lot of the state can be consolidated / simplified

TODOs for this PR:
- some tracing, the current traces are mostly useless now because of the new flows (this was true even before this PR), still we should record a few timestamps and refactor tracing in a separate PR

### DIFF
--- a/crates/api/src/auctioneer/submit_block.rs
+++ b/crates/api/src/auctioneer/submit_block.rs
@@ -98,15 +98,16 @@ impl<A: Api> Context<A> {
         let submission = match submission {
             Submission::Full(full) => full,
             Submission::Dehydrated(dehydrated) => {
+                let max_blobs_per_block = self.chain_info.max_blobs_per_block();
                 let (payload, tx_cache_hits, blob_cache_hits) =
-                    dehydrated.hydrate(&mut self.hydration_cache)?;
+                    dehydrated.hydrate(&mut self.hydration_cache, max_blobs_per_block)?;
 
                 HYDRATION_CACHE_HITS
                     .with_label_values(&["transaction"])
                     .inc_by(tx_cache_hits as u64);
                 HYDRATION_CACHE_HITS.with_label_values(&["blob"]).inc_by(blob_cache_hits as u64);
 
-                payload.validate_payload_ssz_lengths()?;
+                payload.validate_payload_ssz_lengths(max_blobs_per_block)?;
                 payload
             }
         };
--- a/crates/api/src/auctioneer/worker.rs
+++ b/crates/api/src/auctioneer/worker.rs
@@ -196,7 +196,7 @@ impl Worker {
                 trace.signature = Some(utcnow_ns());
             }
 
-            payload.validate_payload_ssz_lengths()?;
+            payload.validate_payload_ssz_lengths(self.chain_info.max_blobs_per_block())?;
             (Submission::Full(payload), merging_data)
         };
 
--- a/crates/api/src/auctioneer/get_payload.rs
+++ b/crates/api/src/auctioneer/get_payload.rs
@@ -3,8 +3,9 @@ use std::sync::Arc;
 use alloy_primitives::B256;
 use helix_common::GetPayloadTrace;
 use helix_types::{
-    BeaconBlockBodyElectra, BeaconBlockElectra, GetPayloadResponse, PayloadAndBlobs,
-    SignedBeaconBlock, SignedBeaconBlockElectra, SignedBlindedBeaconBlock, VersionedSignedProposal,
+    BeaconBlockBodyElectra, BeaconBlockBodyFulu, BeaconBlockElectra, BeaconBlockFulu,
+    GetPayloadResponse, PayloadAndBlobs, SignedBeaconBlock, SignedBeaconBlockElectra,
+    SignedBeaconBlockFulu, SignedBlindedBeaconBlock, VersionedSignedProposal,
 };
 use tokio::sync::oneshot;
 use tracing::{info, warn};
@@ -27,12 +28,12 @@ impl<A: Api> Context<A> {
     ) {
         if self.bid_slot != payload.slot {
             warn!(curr =% self.bid_slot, received =% payload.slot, "received gossiped payload for wrong slot");
-            return
+            return;
         }
 
         if slot_data.proposer_pubkey() != &payload.proposer_pub_key {
             warn!(curr =% slot_data.proposer_pubkey(), received =% payload.proposer_pub_key, "received gossiped payload for wrong proposer");
-            return
+            return;
         }
 
         self.payloads.insert(
@@ -152,7 +153,6 @@ impl<A: Api> Context<A> {
             SignedBlindedBeaconBlock::Bellatrix(_) |
             SignedBlindedBeaconBlock::Capella(_) |
             SignedBlindedBeaconBlock::Deneb(_) |
-            SignedBlindedBeaconBlock::Fulu(_) |
             SignedBlindedBeaconBlock::Gloas(_) => {
                 Err(ProposerApiError::UnsupportedBeaconChainVersion)
             }
@@ -219,6 +219,83 @@ impl<A: Api> Context<A> {
 
                 // TODO: avoid this endless cloning
                 let signed_block = SignedBeaconBlock::Electra(inner).into();
+                let to_broadcast = VersionedSignedProposal {
+                    signed_block,
+                    kzg_proofs: local.blobs_bundle.proofs.clone(),
+                    blobs: local.blobs_bundle.blobs.clone(),
+                };
+                let to_proposer = GetPayloadResponse {
+                    version: slot_data.current_fork,
+                    metadata: Default::default(),
+                    data: local,
+                };
+
+                Ok((to_proposer, to_broadcast))
+            }
+            SignedBlindedBeaconBlock::Fulu(blinded_block) => {
+                // validate
+                // TODO: we should already have the header, as we served it in "get_header"
+                // NOTE: not if it comes via gossip, just implement the check manually
+                let local_execution_payload_header = local
+                    .execution_payload
+                    .to_header(None)
+                    .to_lighthouse_fulu_header()
+                    .map_err(ProposerApiError::SszError)?;
+
+                let block = &blinded_block.message;
+                let body = &block.body;
+                let provided_header = &body.execution_payload.execution_payload_header;
+
+                if &local_execution_payload_header != provided_header {
+                    return Err(ProposerApiError::BlindedBlockAndPayloadHeaderMismatch);
+                }
+
+                let local_kzg_commitments = &local.blobs_bundle.commitments;
+
+                if !local_kzg_commitments.iter().eq(body.blob_kzg_commitments.iter().map(|p| p.0)) {
+                    return Err(ProposerApiError::BlobKzgCommitmentsMismatch);
+                }
+
+                // unblind
+
+                let signature = blinded_block.signature.clone();
+
+                if body.blob_kzg_commitments.len() != local.blobs_bundle.blobs.len() {
+                    return Err(ProposerApiError::BlindedBlobsBundleLengthMismatch);
+                }
+
+                let local_execution_payload = local
+                    .execution_payload
+                    .to_lighthouse_fulu_payload()
+                    .map_err(ProposerApiError::SszError)?;
+
+                let inner = SignedBeaconBlockFulu {
+                    message: BeaconBlockFulu {
+                        slot: block.slot,
+                        proposer_index: block.proposer_index,
+                        parent_root: block.parent_root,
+                        state_root: block.state_root,
+                        body: BeaconBlockBodyFulu {
+                            randao_reveal: body.randao_reveal.clone(),
+                            eth1_data: body.eth1_data.clone(),
+                            graffiti: body.graffiti,
+                            proposer_slashings: body.proposer_slashings.clone(),
+                            attester_slashings: body.attester_slashings.clone(),
+                            attestations: body.attestations.clone(),
+                            deposits: body.deposits.clone(),
+                            voluntary_exits: body.voluntary_exits.clone(),
+                            sync_aggregate: body.sync_aggregate.clone(),
+                            execution_payload: local_execution_payload.into(),
+                            bls_to_execution_changes: body.bls_to_execution_changes.clone(),
+                            blob_kzg_commitments: body.blob_kzg_commitments.clone(),
+                            execution_requests: body.execution_requests.clone(),
+                        },
+                    },
+                    signature,
+                };
+
+                let signed_block = SignedBeaconBlock::Fulu(inner).into();
+
                 let to_broadcast = VersionedSignedProposal {
                     signed_block,
                     kzg_proofs: local.blobs_bundle.proofs.clone(),


crates/api/src/auctioneer/bid_sorter.rs
```diff
@@ -0,0 +1,342 @@
...
+            merging_preferences,
+        }: BidSorterMessage,
+    ) {
+        let recv_ns = utcnow_ns();
+
+        // move this out
+        if self.curr_bid_slot != slot {
+            self.local_telemetry.past_subs += 1;
+            return;
+        }
```
gd-0: does this not contain validated non-optimistic submissions? why are we dropping all demoted keys?

crates/api/src/auctioneer/bid_sorter.rs
```diff
@@ -0,0 +1,342 @@
...
+
+    pub fn best_mergeable(&self) -> Option<BuilderBid> {
+        let curr = self.curr_bid.as_ref()?;
+        let (_, preferences) = self.headers.get(&curr.1.on_receive_ns)?;
+
+        if preferences.allow_appending {
+            Some(curr.2.clone())
+        } else {
+            None
+        }
```
gd-0: up that syntax game

```suggestion
        preferences.allow_appending.then(|| curr.2.clone())
```

crates/api/src/auctioneer/get_header.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        let Some(bid) = self.bid_sorter.get_header() else {
+            return Err(ProposerApiError::NoBidPrepared);
+        };
+
+        // TODO: this check is proably useless
+        if bid.value == U256::ZERO {
+            return Err(ProposerApiError::BidValueZero);
+        }
```
gd-0: yeah think we remove tbh. proposer can just check it

crates/api/src/auctioneer/handle.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl AuctioneerHandle {
+    pub fn new(
+        worker: crossbeam_channel::Sender<WorkerJob>,
+        auctioneer: crossbeam_channel::Sender<Event>,
+    ) -> Self {
+        Self { worker, auctioneer }
+    }
+
+    #[allow(clippy::result_unit_err)]
```
gd-0: what's up with this error crime? why not just return the worker error?

crates/api/src/auctioneer/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+                    withdrawals_root,
+                    sequence,
+                    trace,
+                    res_tx,
+                    slot_data,
+                );
+
+                if let Some(state) = Self::maybe_start_broacasting(ctx, slot_data) {
+                    *self = state;
+                }
```
gd-0, line 277: am I missing something? why would we start broadcasting if we get a new block submission while sorting? doesn't this happen for every new bid??

crates/api/src/auctioneer/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+                } else {
+                    // we may still receive the payload from builder / gossip later
+                    ctx.pending_payload =
+                        Some(PendingPayload { block_hash, blinded, trace, res_tx });
+                }
+            }
+
+            // sim result
+            (State::Sorting(_), Event::SimResult(result)) => {
+                ctx.sort_simulation_result(&result);
```
gd-0: do we call this twice for optimistic submissions? i.e., 
1) when we first recv
2) when we get their sim result back

crates/api/src/auctioneer/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+            (State::Broadcasting { .. }, Event::GetHeader { res_tx, .. }) => {
+                let _ = res_tx.send(Err(ProposerApiError::DeliveringPayload));
+            }
+
+            // duplicate get_payload, proposer equivocating?
+            (
+                State::Broadcasting { slot_data: slot_ctx, block_hash },
+                Event::GetPayload { blinded, block_hash: new_block_hash, res_tx, .. },
+            ) => {
+                if slot_ctx.bid_slot == blinded.slot() || *block_hash == new_block_hash {
```
gd-0: Think this should be AND
```suggestion
                if slot_ctx.bid_slot == blinded.slot() && *block_hash == new_block_hash {
```

crates/api/src/auctioneer/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+                        got.block_hash =% payload.execution_payload.execution_payload.block_hash,
+                        got.slot = payload.slot,
+                        got.pubkey =% &payload.proposer_pub_key,
+                        "mismatch in broadcasting / gossip payload")
+                }
+            }
+
+            // submission unregistered
+            (State::Slot { .. }, Event::Submission { res_tx, .. }) => {
+                let _ = res_tx.send(Err(BuilderApiError::ProposerDutyNotFound));
```
gd-0, line 389: know it's unrelated to this PR but we should change this error message to: `ProposerNotRegistered` imo. much clearer than ProposerDutyNotFound

crates/api/src/auctioneer/submit_block.rs
```diff
@@ -0,0 +1,214 @@
...
+        bid_sorter::BidSorterMessage,
+        context::Context,
+        simulator::{manager::SimulationResult, BlockSimRequest, SimulatorRequest},
+        types::{SlotData, Submission, SubmissionResult},
+    },
+    builder::error::BuilderApiError,
+    Api,
+};
+
+impl<A: Api> Context<A> {
```
gd-0, line 25: not the biggest fan of splitting up a struct like `Context` accross multiple files. Also not even in a `context` mod

crates/api/src/auctioneer/submit_block.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        let builder_info = self.builder_info(submission.builder_public_key());
+
+        self.validate_submission(
+            &submission,
+            &withdrawals_root,
+            sequence,
+            &builder_info,
+            slot_data,
+        )?;
```
gd-0, line 125: what do we get from seperating this fn into its own file?

crates/api/src/auctioneer/submit_block.rs
```diff
@@ -0,0 +1,214 @@
...
+        submission: SignedBidSubmission,
+        submission_trace: SubmissionTrace,
+        optimistic_version: OptimisticVersion,
+        res_tx: Option<oneshot::Sender<SubmissionResult>>,
+        slot_data: &SlotData,
+        merging_preferences: BlockMergingPreferences,
+    ) {
+        // TODO: pass this from previous step
+        let is_top_bid = self.bid_sorter.is_top_bid(&submission);
+        let inclusion_list = slot_data.il.clone();
```
gd-0, line 151: is the IL Arc'd ?

crates/api/src/auctioneer/submit_block.rs
```diff
@@ -0,0 +1,214 @@
...
+        );
+
+        // send to sim_manager queue, use another oneshot or simply the sim results?
+        let req = SimulatorRequest {
+            request,
+            on_receive_ns: submission_trace.receive,
+            is_top_bid,
+            is_optimistic: optimistic_version.is_optimistic(),
+            res_tx,
+            submission: submission.clone(),
```
gd-0, line 165: can't we Arc the submission both here and below for the DB? assuming both just get serialised

crates/api/src/auctioneer/worker.rs
```diff
@@ -0,0 +1,217 @@
...
+
+                        let message = Event::Submission {
+                            // TODO: move this to auctioneer, avoid clones
+                            submission: submission.clone(),
+                            merging_preferences,
+                            withdrawals_root,
+                            sequence,
+                            trace,
+                            res_tx,
+                        };
```
gd-0: we should avoid cloning the submission in the hotpath given we only need the data below if block merging is not enabled. just put this behind the block merging if:

```rust
if !self.config.block_merging_config.is_enabled {
   // send without cloning
} else {
  // send with cloning
  // other stuff in this if
```

crates/types/src/hydration.rs
```diff
@@ -19,11 +21,20 @@ pub enum DehydratedBidSubmission {
...
-            DehydratedBidSubmission::Electra(dehydrated_bid_submission_electra) => {
-                &dehydrated_bid_submission_electra.message.builder_pubkey
-            }
+            DehydratedBidSubmission::Electra(s) => &s.message.builder_pubkey,
+        }
+    }
+
+    pub fn withdrawal_root(&self) -> B256 {
+        match self {
+            DehydratedBidSubmission::Electra(s) => s.execution_payload.withdrawals.tree_hash_root(),
```
gd-0, line 48: same as with the tx root. dehydrated submissions should include the withdrawals root. we can then add this as a demotiuon condition during optimistic validation

crates/housekeeper/src/inclusion_list/service.rs
```diff
@@ -12,66 +12,83 @@ use helix_network::RelayNetworkManager;
...
+        match self.db.save_inclusion_list(&inclusion_list, head_slot, &parent_hash, &pub_key).await
+        {
             Ok(_) => {
-                info!(head_slot = slot, "Saved inclusion list to postgres");
+                info!(head_slot = head_slot, "Saved inclusion list to postgres");
             }
             Err(err) => {
                 error!(
-                    head_slot = slot,
+                    head_slot = head_slot,
```
gd-0: if the name matches no need for = in tracing:
```suggestion
                    head_slot,
```

crates/housekeeper/src/chain_event_updater.rs
```diff
@@ -38,35 +48,41 @@ pub struct SlotUpdate {
...
+
 /// Manages the update of head slots and the fetching of new proposer duties.
 pub struct ChainEventUpdater {
     head_slot: u64,
     known_payload_attributes: HashMap<(B256, Slot), PayloadAttributesEvent>,
-
     proposer_duties: Vec<BuilderGetValidatorsResponseEntry>,
-
-    auctioneer: Arc<LocalCache>,
+    local_cache: Arc<LocalCache>,
```
gd-0, line 65: Arcing a "Local" cache seems wrong :sweat_smile:

crates/common/src/config.rs
```diff
@@ -60,6 +58,20 @@ pub struct RelayConfig {
...
     pub is_local_dev: bool,
+    #[serde(default)]
+    pub auctioneer: AuctioneerConfig,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct AuctioneerConfig {
+    #[serde(default = "default_usize::<8>")]
+    pub worker_threads: usize,
+}
```
gd-0: imo wouldn't put this as part of a seperate config. the whole thing is really the auctioneer now. would set it as the threads needed for the worker stage in the main config

crates/api/src/auctioneer/bid_sorter.rs
```diff
@@ -0,0 +1,342 @@
...
+            slot: self.curr_bid_slot,
+            block_number: h.header.block_number,
+            block_hash: h.header.block_hash,
+            parent_hash: h.header.parent_hash,
+            builder_pubkey,
+            fee_recipient: h.header.fee_recipient,
+            value: bid.value,
+        }
+        .as_ssz_bytes()
+        .into();
```
gd-0, line 252: small optimisation but given we know the size of TopBidUpdate we can prealloc the vec each time and use `ssz_append`.
```suggestion
        let mut buf = Vec::with_capacity(size_of::<TopBidUpdate>);
        TopBidUpdate {
            timestamp: utcnow_ms(),
            slot: self.curr_bid_slot,
            block_number: h.header.block_number,
            block_hash: h.header.block_hash,
            parent_hash: h.header.parent_hash,
            builder_pubkey,
            fee_recipient: h.header.fee_recipient,
            value: bid.value,
        }.ssz_append(&mut buf);
        let top_bid_update = buf.into();

```

crates/api/src/auctioneer/bid_sorter.rs
ltitanb: assume you mean the part below there
```rust
if self.demotions.contains(&builder_pubkey) {
            self.local_telemetry.demoted_subs += 1;
            return;
}
```
that was the previous behaviour as well so I haven't changed that, if an optimistic builder is demoted we would drop all bids for that slot (not future slots)

crates/api/src/auctioneer/handle.rs
ltitanb: The only error there is just that the channel is full, and clippy was complaining that the error variant was very large `TrySendError<Event>`  apparently is large if `Event` is large: https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
instead of boxing it , i just used `()` since we dont need to match the error, alternative is to use an `Option` but result was more expressive i thought

crates/api/src/auctioneer/mod.rs
ltitanb: if we received a `get_payload` and did not have the block, we save it in `pending_payload`. When we receive submissions or gossiped payloads we check if we have any pending payload and if we just received the block. This is similar to the previous behaviour of trying to keep fetching the payload repeatedly in `get_payload`

https://github.com/gattaca-com/helix/blob/lt/new-auctioneer/crates/api/src/auctioneer/mod.rs#L447-L451

crates/api/src/auctioneer/mod.rs
ltitanb: nope, for optimistic submission we don't do 2: https://github.com/gattaca-com/helix/blob/lt/new-auctioneer/crates/api/src/auctioneer/submit_block.rs#L70-L92

crates/api/src/auctioneer/mod.rs
ltitanb: yep i left a general TODO to refactor errors as would think many of them are unused now

crates/housekeeper/src/chain_event_updater.rs
ltitanb: agreed, should probably be named `SharedCache`

crates/api/src/auctioneer/bid_sorter.rs
ltitanb: seems like it's different actually, 
```
assert_eq!(top_bid.as_ssz_bytes().len(), std::mem::size_of::<TopBidUpdate>());
assertion `left == right` failed
  left: 188
 right: 192
```
but yep can add that as a constant

crates/api/src/auctioneer/bid_sorter.rs
gd-0: ok this is wrong. we just demote immediately then process all bids after that non-optimistically

crates/api/src/auctioneer/handle.rs
gd-0: we should probably communicate that it's full? maybe a custom error here is better

crates/api/src/auctioneer/mod.rs
gd-0: so we never get a sim result event for optimistic sims??

crates/api/src/auctioneer/bid_sorter.rs
gd-0: ah yeah will need offsets

crates/api/src/auctioneer/mod.rs
ltitanb: we do https://github.com/gattaca-com/helix/blob/lt/new-auctioneer/crates/api/src/auctioneer/submit_block.rs#L47-L61 , but wont sort it again

crates/api/src/auctioneer/submit_block.rs
ltitanb: think it's a lot cleaner than in a single file , as it's split by request type. Similar to how we did it in the APIs

crates/api/src/auctioneer/submit_block.rs
ltitanb: keep the validation logic contained since there are a bunch of different checks

crates/api/src/auctioneer/submit_block.rs
ltitanb: it's `Bytes` underneath so basically an Arc, there's still a vec that is cloned, can fix that

crates/api/src/auctioneer/bid_sorter.rs
ltitanb: i removed that check for now, now that the demotions and the sorter is in the same thread , a lot of these checks can be consolidated. To be done in a follow up PR

crates/api/src/auctioneer/submit_block.rs
ltitanb: it's one of the TODOs i mentioned above, most state can be simplified and a lot of cloning can be avoided. I can do this one in this PR if you prefer

crates/api/src/auctioneer/mod.rs
gd-0: so this is basically a tick to check it? if so can you add a comment here. was a bit confusing

crates/api/src/auctioneer/submit_block.rs
gd-0: but we have calls accross files for the same context. reasoning about the flow is much harder. 
Not a blocker for this PR but we should have a general discussion about this in the coding style chat.

crates/api/src/auctioneer/submit_block.rs
gd-0: if you've got a list of all of these optimisations lets do in seperate PRs.

crates/common/src/utils.rs
```diff
@@ -200,3 +200,7 @@ pub fn avg_duration(duration: Duration, count: u32) -> Option<Duration> {
         None
     }
 }
+
+pub fn pin_thread_to_core(core: usize) -> bool {
+    core_affinity::set_for_current(core_affinity::CoreId { id: core })
+}
```
gd-0, line 209: what does this wrapper fn actually add?

crates/api/src/auctioneer/decoder.rs
```diff
@@ -79,11 +80,46 @@ impl SubmissionDecoder {
...
+            Encoding::Ssz => {
+                const BUILDER_PUBKEY_OFFSET: usize = 8 + /* slot */
+                32 + /* parent_hash */
+                32; /* block_hash */
+
+                if buf.len() < BUILDER_PUBKEY_OFFSET + BlsPublicKeyBytes::len_bytes() {
+                    return Err(BuilderApiError::PayloadDecode);
+                }
+
+                let pubkey = unsafe {
```
0w3n-d, line 111: While the bounds check above does seem to be enough to make this safe, I think anytime we are passing user controlled data directly to unsafe rust like this we should have a bunch of unit tests to make sure there's no bugs. Maybe even fuzz the input.

ltitanb: added a unit test but don't think we need fuzzing? the pubkey is just a wrapper around `[u8; 48]`  and the check is quite simple

PR #266, opened 2025-09-29
upgrades relay to fulu

0w3n-d:


crates/api/src/builder/simulator/client.rs
```diff
@@ -64,10 +74,19 @@ impl SimulatorClient {
             headers.insert("X-High-Priority", HeaderValue::from_static("true"));
         };
 
+        let mut sim_method = &self.sim_method_v4;
+
+        if request.blobs_bundle.is_some() &&
+            request.blobs_bundle.as_ref().unwrap().proofs.len() >
+                request.blobs_bundle.as_ref().unwrap().commitments.len()
```
gd-0, line 81: this seems like maybe not the most robust way to communicate the method. givent he submission is an enum can't we use that to attatch the version?

crates/api/src/gossiper/types/broadcast_get_payload.rs
```diff
@@ -44,7 +44,7 @@ fn decode_ssz_signed_blinded_beacon_block(
     }
 
     if let Ok(signed_blinded_beacon_block) =
-        SignedBlindedBeaconBlock::from_ssz_bytes_by_fork(bytes, ForkName::Electra)
+        SignedBlindedBeaconBlock::from_ssz_bytes_by_fork(bytes, ForkName::Fulu)
```
gd-0: is this just for the default case in case no fork_name is sent? when would there ever be no fork_name anyway? whaqt's the point of this default

crates/api/src/proposer/get_payload.rs
```diff
@@ -294,7 +373,7 @@ impl<A: Api> ProposerApi<A> {
             (trace_clone, failed_publishing)
         });
 
-        if !is_trusted_proposer {
+        if !is_trusted_proposer && matches!(api_version, ProposerApiVersion::V1) {
```
gd-0, line 380: why is it safe to skip trusted_proposer checks if on V2?

crates/api/src/proposer/get_payload.rs
```diff
@@ -621,6 +701,35 @@ fn validate_block_equality(
...
+            let provided_kzg_commitments = body
+                .blob_kzg_commitments()
+                .map_err(|_| ProposerApiError::BlobKzgCommitmentsMismatch)?;
+
+            if !local_kzg_commitments.iter().eq(provided_kzg_commitments.iter().map(|p| p.0)) {
+                return Err(ProposerApiError::BlobKzgCommitmentsMismatch);
+            }
+        }
+        BlindedPayloadRef::Fulu(blinded_payload) => {
+            info!(
```
gd-0: if we want to make this log info we should also make the above version info instead of debug

crates/api/src/builder/api.rs
```diff
@@ -124,7 +124,7 @@ impl<A: Api> BuilderApi<A> {
         });
 
         let (hydration_tx, hydration_rx) = mpsc::channel(10_000);
-        hydration::spawn_hydration_task(hydration_rx);
+        hydration::spawn_hydration_task(hydration_rx, chain_info.max_blobs_per_block());
```
ltitanb: this can change across the fork no? shouldn't we recompute it for every slot?

crates/api/src/builder/simulator/client.rs
ltitanb, line 81: let's add a get_fork() on the `BlobsBundle` instead so it's a bit cleaner?

crates/api/src/builder/simulator/rpc_simulator.rs
ltitanb, line 1: this seems the old simulator, why do we need it here?

crates/api/src/gossiper/types/broadcast_get_payload.rs
ltitanb: won't this break before the fork?

crates/types/src/hydration.rs
```diff
@@ -57,15 +73,25 @@ pub struct DehydratedBidSubmissionElectra {
...
+    execution_payload: ExecutionPayload,
+    blobs_bundle: DehydratedBlobs,
+    execution_requests: Arc<ExecutionRequests>,
+    signature: BlsSignatureBytes,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode)]
 struct DehydratedBlobs {
-    proofs: Vec<KzgProof>,
+    commitments: Vec<KzgCommitment>,
```
gd-0, line 88: are commitments also unique per proof?

crates/api/src/gossiper/types/broadcast_payload.rs
```diff
@@ -52,8 +52,7 @@ fn decode_ssz_payload_and_blobs(
         return PayloadAndBlobs::from_ssz_bytes_by_fork(bytes, fork_name);
     }
 
-    if let Ok(payload_and_blobs) = PayloadAndBlobs::from_ssz_bytes_by_fork(bytes, ForkName::Electra)
-    {
+    if let Ok(payload_and_blobs) = PayloadAndBlobs::from_ssz_bytes_by_fork(bytes, ForkName::Fulu) {
```
ltitanb: same here

crates/api/src/proposer/types.rs
```diff
@@ -72,6 +72,55 @@ pub fn unblind_beacon_block(
...
+                .execution_payload
+                .clone()
+                .to_lighthouse_fulu_payload()
+                .map_err(ProposerApiError::SszError)?;
+
+            let blobs_bundle = &versioned_execution_payload.blobs_bundle;
+
+            if body.blob_kzg_commitments.len() != blobs_bundle.blobs.len() {
+                return Err(ProposerApiError::BlindedBlobsBundleLengthMismatch);
+            }
```
ltitanb, line 95: just double checking that this is expected to be of the same length?

crates/api/src/gossiper/types/broadcast_get_payload.rs
ltitanb: think this was during a transition time when we added the fork field to the .proto

gd-0: if that's the case then we should remove

crates/api/src/proposer/get_payload.rs
0w3n-d: Because V2 doesn't return any data back to the proposer

crates/api/src/proposer/types.rs
0w3n-d: yep

crates/types/src/hydration.rs
0w3n-d: Previously we could the proof as key for a BlobItem because there was one proof per blob. But that's changed with peer das. Now each blob will be associated with 128 proofs.

But each blob is still associated with only one commitment so we can use the commitment as the key for the BlobItems in the cache now instead.

crates/api/src/builder/simulator/client.rs
0w3n-d: Will fix in follow up PR

crates/api/src/gossiper/types/broadcast_get_payload.rs
0w3n-d: done

crates/api/src/proposer/get_payload.rs
0w3n-d: yeah this should also have been debug. Changed

crates/api/src/builder/api.rs
0w3n-d: changed

crates/api/src/builder/simulator/client.rs
0w3n-d: with address in follow up PR

crates/api/src/builder/simulator/rpc_simulator.rs
0w3n-d: removed

crates/api/src/gossiper/types/broadcast_get_payload.rs
0w3n-d: removed

crates/api/src/gossiper/types/broadcast_payload.rs
0w3n-d: removed

PR #268, opened 2025-10-02
refactor registrations

ltitanb:
refactor registrations to use workers + avoid a bunch of cloning


crates/api/src/auctioneer/worker.rs
```diff
@@ -250,3 +279,40 @@ fn verify_signed_blinded_block_signature(
...
+/// Validates the timestamp in a `SignedValidatorRegistration` message.
+///
+/// - Ensures the timestamp is not too early (before genesis time)
+/// - Ensures the timestamp is not too far in the future (current time + 10 seconds).
+fn validate_registration_time(
+    chain_info: &ChainInfo,
+    registration: &SignedValidatorRegistration,
+) -> Result<(), ProposerApiError> {
+    let registration_timestamp = registration.message.timestamp;
+    let registration_timestamp_upper_bound = utcnow_sec() + 10;
```
gd-0, line 303: we validate batches of registrations in a spin loop. should probably just call utc_now_sec once and pass `now` each time this fn is called.

crates/api/src/proposer/register.rs
```diff
@@ -116,129 +110,101 @@ impl<A: Api> ProposerApi<A> {
...
-        let mut handles = Vec::with_capacity(registrations.len());
-
         let mut unknown_registrations = 0;
         let mut update_not_required = 0;
 
-        for registration in registrations {
-            let proposer_api_clone = proposer_api.clone();
-            let start_time = Instant::now();
+        let registrations_to_check: Vec<_> = {
+            let known_validators_guard = proposer_api.db.known_validators_cache().read();
```
0w3n-d, line 118: I wonder if we should move this cache out of the db at some point? It probably makes sense for it to be kept in memory on a reg instance. Never pull from the db and kept updated by a process that runs on a reg instance rather than the housekeeper.

crates/database/src/lib.rs
```diff
@@ -29,12 +32,25 @@ pub async fn start_db_service(
...
+        // housekeeper already runs in the submission instance so no need to refresh there
+        let should_refresh_cache = !config.is_submission_instance;
         tokio::spawn({
             let known_validators_loaded = known_validators_loaded.clone();
             let postgres_db = postgres_db.clone();
             async move {
                 postgres_db.load_known_validators().await;
                 known_validators_loaded.store(true, Ordering::Relaxed);
+
+                if should_refresh_cache {
```
0w3n-d, line 44: I think loading from db on startup likely makes sense to so we don't need to wait for it do download from the beacon client before we can start processing registration, but I wonder for the refresh would it make sense for the reg instance to just fetch it's own know validators from the beacon client? Would could even just make the know validators stuff a reg box only job as I don't think the submission instance needs it and the data api definitely doesn't.

crates/database/src/postgres/postgres_db_service_tests.rs
```diff
@@ -123,113 +121,113 @@ mod tests {
...
-            });
-        }
-
-        let result = db_service.set_proposer_duties(proposer_duties).await;
-        assert!(result.is_ok());
-
-        let result = db_service.get_proposer_duties().await;
-        assert!(result.is_ok());
-    }
+    // #[tokio::test]
```
0w3n-d: When commenting blocks of code out like this I think we should either just delete it instead or add a TODO to fix the test or whatever it is we plan do with the commented out code in the future

PR #269, opened 2025-10-03
adds support for BlobsBundleV2 for fulu

0w3n-d:


crates/types/src/blobs.rs
```diff
@@ -83,12 +226,180 @@ impl BlobsBundleV1 {
...
+    }
+
+    /// Create with pre-allocated capacity
+    pub fn with_capacity_v1(capacity: usize) -> Self {
+        BlobsBundle::V1(Arc::new(BlobsBundleV1::with_capacity(capacity)))
+    }
+
+    pub fn with_capacity_v2(capacity: usize) -> Self {
+        BlobsBundle::V2(Arc::new(BlobsBundleV2::with_capacity(capacity)))
+    }
```
gd-0: how are these ever used if we insta arc?

crates/types/src/blobs.rs
```diff
@@ -83,12 +226,180 @@ impl BlobsBundleV1 {
...
+}
+
+impl Default for BlobsBundle {
+    fn default() -> Self {
+        Self::V1(Arc::new(BlobsBundleV1::default()))
+    }
+}
+
+pub enum BlobsBundleMut {
```
gd-0: what usecase do we have for wanting a mutable blobs bundle??

Seems to me like a pattern we should never support. If we are ever rebuilding blobs bundles from deltas etc we should not Arc the entire bundle but instead, just arc the induvidual `Blobs` or cells. or use Bytes

crates/types/src/blobs.rs
0w3n-d: Looks like these aren't actually used, must have been part of one of my earlier iterations, I'll remove them

crates/types/src/blobs.rs
0w3n-d: It's used in the block merging stuff to append blobs. It's horrible. But I added because I changed the definition of BlobsBundle to: 

`pub enum BlobsBundle {
    V2(Arc<BlobsBundleV2>),
    V1(Arc<BlobsBundleV1>),
}`

Also, horrible but I wanted to be able to call blobs_bundle() on a `SignedBidSubmission` without cloning the blobs

gd-0: if we make the inner blob a ref then this is no issue. this doesn't solve any of the copying because you need to copy out of the arc to get to this anyway. 

Instead of making the BlobsBundle { variant: Arc<Type>}
we can make the Type have refs and deser straight into the ref

crates/types/src/blobs.rs
gd-0: also some comment crime on these :wink:

crates/types/src/blobs.rs
0w3n-d: OK I'll try that.

Currently I deser straight to `Arc<BlobsBundleV1>` and only wrap it in the BlobsBundle enum when blobs_bundle() get's called. I only copy out of the Arc when creating the BlobsBundleMut in the merge block stuff

crates/types/src/blobs.rs
0w3n-d: remove

crates/types/src/blobs.rs
0w3n-d: BlobsBundleMut is gone now

crates/types/src/blobs.rs
```diff
@@ -83,12 +226,130 @@ impl BlobsBundleV1 {
...
+        &mut self,
+        commitment: KzgCommitment,
+        proofs: &[KzgProof],
+        blob: Blob,
+    ) -> Result<(), BlobsError> {
+        match self {
+            BlobsBundle::V2(bundle) => {
+                bundle.commitments.push(commitment).map_err(|_| BlobsError::BundleTooLarge {
+                    got: bundle.blobs.len() + 1,
+                    max: 6,
```
ltitanb: where is this 6 coming from?

crates/types/src/bid_submission.rs
```diff
@@ -510,28 +627,88 @@ pub enum BidValidationError {
...
+        let s = test_decode_json::<SignedBidSubmission>(data_json);
+
+        let blobs_empty = match &s {
+            SignedBidSubmission::Electra(electra) => electra.blobs_bundle.blobs().is_empty(),
+            SignedBidSubmission::Fulu(fulu) => fulu.blobs_bundle.blobs().is_empty(),
+        };
+
+        if blobs_empty {
+            // When blobs are empty, we can't distinguish variants reliably
+            assert!(matches!(s, SignedBidSubmission::Electra(_) | SignedBidSubmission::Fulu(_)));
```
ltitanb: not entirely clear how we're dealing with this case in code? dont we risk to get the wrong variant?

0w3n-d: Yeah apart from blobs the the two bid submissions are identical and blobs are optional so it there's no way to handle this in the code. It will get decoded to the fulu variant and treated as such but should have no impact on how it get's processed as there's no blobs that can fail validation.

crates/types/src/blobs.rs
0w3n-d: Fixed

crates/types/src/bid_submission.rs
ltitanb: This seems a footgun though, as we may call `.fork_name` later and get the wrong one. We could have a `.maybe_upgrade_to_fulu` call just after it's decoded so we know it's correct from then onwards, would still keep Electra as the first variant so the if they can be both it will be Electra

crates/api/src/builder/api.rs
```diff
@@ -130,16 +130,27 @@ fn blobs_bundle_to_hashmap(
...
+                    commitment: *commitment,
+                    proof: proofs[0],
+                    blob: blob.clone(),
+                }),
+            ),
+            BlobsBundleVersion::V2 => (
+                versioned_hash,
+                BlobWithMetadata::V2(BlobWithMetadataV2 {
+                    commitment: *commitment,
+                    proofs: proofs.to_vec(),
```
gd-0, line 150: not sure how big proofs can get but is it is big then this could be a decent alloc. might be worth Arcing this as well?

crates/types/src/bid_submission.rs
0w3n-d: OK fair point. Added.

crates/api/src/builder/api.rs
0w3n-d: It's 6 KB. I can look at optimizing this in a follow up PR

PR #271, opened 2025-10-05
lazy builder bid

ltitanb:
messed up a bit the commits so ended up combining 2 PRs..

main change here is the refactor of builder bids for get_header. Previously we would compute the builder bid for every bid received. This is not ideal because it involves computing the tx root which is quite slow. Moreover, only a small subset of submissions are served in get_header, so we would delay the top bid update for nothing. Now we compute the builder bid  lazily in get_header, this gets the submission -> top bid latency to around 500us when tested locally. the downside is that if we were to serve the same header twice we would re compute the root twice 

what got mixed in is an initial refactor of metrics + some cleanup

### DIFF
--- a/crates/api/src/proposer/get_header.rs
+++ b/crates/api/src/proposer/get_header.rs
@@ -125,7 +125,10 @@ impl<A: Api> ProposerApi<A> {
 
             let sleep_time_ms = std::cmp::min(max_sleep_time, target_sleep_time);
 
-            let sleep_time = Duration::from_millis(sleep_time_ms);
+            // leave additional time to resign the compute and sign the builder bid
+            let sleep_time_adj = sleep_time_ms.saturating_sub(5);
+
+            let sleep_time = Duration::from_millis(sleep_time_adj);
             let mut get_header_metric = GetHeaderMetric::new(sleep_time);
 
             debug!(target: "timing_games",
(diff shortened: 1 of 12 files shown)


crates/api/src/auctioneer/bid_sorter.rs
```diff
@@ -226,35 +228,35 @@ impl BidSorter {
...
         let _ = self.top_bid_tx.send(top_bid_update);
 
-        self.curr_bid = Some((builder_pubkey, bid, h.clone()));
+        self.curr_bid = Some((builder_pubkey, bid));
 
         if let Some(trace) = trace {
-            trace.top_bid_at = Some((now_ns, is_cancel));
+            if is_optimistic {
+                // this is our "tick to trade" but may be confounded if builder is sending slowly
+                record_submission_step_ns("recv_top_bid", trace.receive, now_ns);
```
gd-0, line 256: not the true "ttt" as we still need to compute the tx root in the hot-path. Just later on. 
We should add providing the tx root as part of the rehydration data or similar asap to remove that

crates/api/src/proposer/get_header.rs
```diff
@@ -125,7 +125,10 @@ impl<A: Api> ProposerApi<A> {
 
             let sleep_time_ms = std::cmp::min(max_sleep_time, target_sleep_time);
 
-            let sleep_time = Duration::from_millis(sleep_time_ms);
+            // leave additional time to resign the compute and sign the builder bid
+            let sleep_time_adj = sleep_time_ms.saturating_sub(5);
```
gd-0, line 129: im assuming that the root latency scales linearly with the num of txs. maybe we can use that as a better estimate. adding a flat 5ms to the hotpath seems painful

ltitanb: will have a proper think when doing the timing games refactor, would need some synchronization from the sorter since we sleep before knowing the header

PR #272, opened 2025-10-05
add receive check

ltitanb:
check the receive timestamp when discarding bids, left the check in the bid sorter because we also need to sort submission after simulation


crates/api/src/auctioneer/validation.rs
```diff
@@ -97,21 +95,35 @@ impl<A: Api> Context<A> {
...
-                // higher sequence number, update
-                *old_seq = new_seq;
+        if let Some((old_receive_ns, maybe_old_seq)) = self.sequence.get_mut(builder) {
+            if new_receive_ns > *old_receive_ns {
+                *old_receive_ns = new_receive_ns
             } else {
-                // stale or duplicated sequence number
-                return Err(BuilderApiError::OutOfSequence { seen: *old_seq, this: new_seq });
+                return Err(BuilderApiError::AlreadyProcessingNewerPayload);
+            }
```
gd-0: why is seq number not taking precedence in the error here? 
maybe i'm not understanding correctly but i assume that this change is to hard override the recv time of blocks to the latest so that if we get seq numbers out of time order we can still use the latest time in the sorter.
Why does this make it ok to return if the new receive is < old receive?

ltitanb: so in the sorter we consider the receive timestamp regardless and simply discard the bid, so it makes sense to discard even before and avoid processing (simulating etc). This is an additional check on top of the sequence number and if seq is present it's still going to be enforced

0w3n-d: Is it possible for seq 1 to arrive later than seq 2 but hit this function in order? So old is seq 1 and new is seq 2 and seq 2 gets rejected by the first check even though would have passed the seq check?

gd-0: should be totally possible right? we take the recv_ts on first packet, then could just have a longer download for one

ltitanb: yes that would be similar to the current behaviour, as the sorter, simulator manager etc only takes into account the receive timestamp to decide when to keep a submission. Even swapping the checks would have the same issue though no? so would need to create a "timestamp" with both sequence and receive_ns and always compare both

PR #307, opened 2025-10-18
fork aware auction

ltitanb:
if slot N-1 is slow to propagate, we may receive two payload attributes updates, one with parent hash N-2 and one with N-1. The proposer will choose which fork to fetch the header for, so the relay needs to hold multiple concurrent auctions for each potential fork. In theory there could be even more than 2 competing forks so I use a map from parent hash -> state. This will add some overhead especially in the sorting, but shouldn't be too large

example on hoodi
```ls
bid sorter telemetry slot=1550986 valid_subs=47 avg_sub_process=Some(3.179µs) fork_report=["parent: 0xc9da18a04d470c0e63df0b0f3b6defb7b58ba7475bf402d12ae1c103498da78e, subs: 8, top_bids: 6", "parent: 0x7bcb0572dfd4a8491880a7e30b7f76e0f7cd1f093b717205db037edf68797a49, subs: 37, top_bids: 14"]
```

### DIFF
--- a/crates/relay/src/api/proposer/get_header.rs
+++ b/crates/relay/src/api/proposer/get_header.rs
@@ -35,7 +35,7 @@ impl<A: Api> ProposerApi<A> {
     /// The function returns a JSON response containing the best bid if found.
     ///
     /// Implements this API: <https://ethereum.github.io/builder-specs/#/Builder/getHeader>
-    #[tracing::instrument(skip_all, err(level = tracing::Level::TRACE), fields(id =% extract_request_id(&headers), slot = params.slot))]
+    #[tracing::instrument(skip_all, err(level = tracing::Level::TRACE), fields(id =% extract_request_id(&headers), slot = params.slot, parent_hash =? params.parent_hash))]
     pub async fn get_header(
         Extension(proposer_api): Extension<Arc<ProposerApi<A>>>,
         Extension(timings): Extension<RequestTimings>,


crates/relay/src/auctioneer/bid_sorter.rs
```diff
@@ -155,60 +251,57 @@ impl BidSorter {
...
             Some((curr_pubkey, curr_bid)) => {
                 if new_bid.value > curr_bid.value {
-                    self.update_top_bid(new_pubkey, new_bid, Some(trace), is_optimistic);
+                    state.update_top_bid(
+                        self.curr_bid_slot,
+                        new_pubkey,
+                        new_bid,
+                        Some(trace),
+                        is_optimistic,
+                        &self.top_bid_tx,
```
0w3n-d, line 264: If we are running multiple auctions but only have this one top bid feed, might builders see a strange looking stream of bids? Will they need to check the parent hash of the incoming top_bid? Or do they already do that?

ltitanb: yep they need to check that, ideally they do this already

0w3n-d (approved):

crates/relay/src/auctioneer/bid_sorter.rs
```diff
@@ -54,22 +54,114 @@ impl BidEntry {
...
+
 pub struct BidSorter {
     /// Sender for ws updates, TopBidUpdate SSZ encoded
     top_bid_tx: tokio::sync::broadcast::Sender<Bytes>,
     /// Head slot + 1
     curr_bid_slot: u64,
-    /// All bid entries for the current slot, used for sorting
-    bids: FxHashMap<BlsPublicKeyBytes, BidEntry>,
+    /// Parent hash -> fork state
+    forks: FxHashMap<B256, ForkState>,
```
gd-0, line 162: doesn't this need to be the full slot coordinate? Could in theory also be viewing the wrong slot/ proposer pubkey i thikn

ltitanb: the slot is validated already via the state of the auctioneer, proposer pubkey is fixed for a given epoch so think the parent hash should be enough

PR #308, opened 2025-10-18
submission version

ltitanb:
encapsulate sequence + receive timestamp checks for builder submissions

### DIFF
--- a/crates/types/src/bid_submission.rs
+++ b/crates/types/src/bid_submission.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{cmp::Ordering, sync::Arc};
 
 use alloy_primitives::{Address, B256, U256};
 use lh_types::{ForkName, SignedRoot, Slot, test_utils::TestRandom};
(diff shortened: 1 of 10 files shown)


crates/types/src/bid_submission.rs
```diff
@@ -633,6 +633,43 @@ impl SignedBidSubmission {
...
+            write!(f, "(recv_ns: {})", self.on_receive_ns)
+        }
+    }
+}
+
+impl Ord for SubmissionVersion {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match (self.sequence, other.sequence) {
+            (Some(a), Some(b)) => a.cmp(&b),
+            _ => self.on_receive_ns.cmp(&other.on_receive_ns),
```
gd-0, line 662: do we not want to still cover the cases where there is a Some/None combo?

ltitanb: dont think so, builder should either set numbers or not for a given slot

gd-0: yep agreed.

PR #311, opened 2025-10-20
remove block hash dedup check

ltitanb:


gd-0 (approved):
logic behind this is just that we rarely get dups and can handle them further down anyway?

Issue #283, opened 2025-10-10
Catch validation error when broadcasting and demote builder


ltitanb:
Because of optimistic submissions and how we handle simulations, a block may be selected for broadcasting without ever being simulated. Leading to errors like this:

```
Block accepted but not processed: {"content-type": "application/json", "server": "Lighthouse/v7.0.0-54f7bc5/x86_64-linux", "content-length": "434", "date": "Wed, 08 Oct 2025 19:26:49 GMT"} body: "{\"code\":202,\"message\":\"ACCEPTED: the object was broadcast to the network without being fully imported to the local database: ExecutionPayloadError(RejectedByExecutionEngine { status: Invalid { latest_valid_hash: None, validation_error: Some(\\\"block hash mismatch: want 0x1c4c5c4070f1f84988d207888845dae06a8b5a36a8302a77c011186ae69190b0, got 0x05ed38849ab9048d072c7d4b4ba077a413700feb72b14210bae67bc25b69927c\\\") } })\",\"stacktraces\":[]}" id="27c07520-05e3-4b53-a52a-2d5783cef461"
```

we should catch this error and demote the builder

tosynthegeek:
Hello @ltitanb I’d like to take up this issue. 

My approach would be to catch this error by checking when the response body contains `ExecutionPayloadError` or specific messages like `"block hash mismatch"`, and then trigger a demotion for the corresponding builder.

tosynthegeek:
> Hi [@tosynthegeek](https://github.com/tosynthegeek) ! great to hear, the specific error we want to catch right now looks like this:
> 
> ```
[...]

PR #315, opened 2025-10-23
builder demotion for block validation errors (#283)

tosynthegeek:
This PR addresses issue #283 by implementing builder demotion for `BlockValidationFailed` errors (e.g., block hash mismatch). The changes ensure that the builder for the invalid block is demoted once per incident, using both cache (`local_cache.demote_builder`) and database (`db.db_demote_builder`) for immediate and persistent effects, respectively.

## Changes
- Added `builder_pubkey: Option<BlsPublicKeyBytes>` to `GetPayloadResultData` to pass the builder's public key from `bid_sorter.get_builder_pub_key`.
- Modified `handle_get_payload` to retrieve `builder_pubkey` from `ForkState` using `parent_hash`.
- Updated `ProposerApi::_get_payload` to demote builders on `BlockValidationFailed` errors after `MultiBeaconClient::publish_block`.

### DIFF
--- a/crates/relay/src/api/proposer/get_payload.rs
+++ b/crates/relay/src/api/proposer/get_payload.rs
@@ -5,7 +5,7 @@ use axum::{Extension, http::HeaderMap, response::IntoResponse};
 use helix_common::{
     Filtering, GetPayloadTrace, RequestTimings,
     api_provider::ApiProvider,
-    beacon::types::BroadcastValidation,
+    beacon::{BeaconClientError, types::BroadcastValidation},
     chain_info::ChainInfo,
     decoder::{Encoding, HEADER_SSZ},
     spawn_tracked,
@@ -24,6 +24,7 @@ use uuid::Uuid;
 
 use super::ProposerApi;
 use crate::{
+    Event,
     api::{
         Api,
         proposer::{CONSENSUS_VERSION_HEADER, error::ProposerApiError},
@@ -369,6 +370,25 @@ impl<A: Api> ProposerApi<A> {
                 )
                 .await
             {
+                if matches!(err, BeaconClientError::BlockValidationFailed(..)) {
+                    let builder_pubkey = bid.builder_pubkey;
+                    let reason = format!("Block validation failed: {err}");
+
+                    let _ = self_clone.auctioneer_handle.send_event(Event::BuilderDemotion {
+                        slot,
+                        builder_pubkey,
+                        block_hash,
+                        reason,
+                    });
+
+                    warn!(
+                        %builder_pubkey,
+                        %block_hash,
+                        slot = %slot,
+                        "BuilderDemotion event sent due to BlockValidationFailed"
+                    );
+                }
+
                 error!(%err, "error publishing block");
                 failed_publishing = true;
             };
--- a/crates/relay/src/auctioneer/types.rs
+++ b/crates/relay/src/auctioneer/types.rs
@@ -465,6 +465,12 @@ pub enum Event {
     GossipPayload(BroadcastPayloadParams),
     SimResult(ValidationResult),
     MergeResult(MergeResult),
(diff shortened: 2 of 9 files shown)


crates/relay/src/auctioneer/bid_sorter.rs
```diff
@@ -224,6 +224,10 @@ impl BidSorter {
         self.forks.get(parent_hash).and_then(|s| s.curr_bid.as_ref().map(|b| b.1.block_hash))
     }
 
+    pub fn get_builder_pub_key(&self, parent_hash: &B256) -> Option<BlsPublicKeyBytes> {
```
ltitanb: the builder pubkey should be already in the "local" payload we have in the payload entry, if not let's add it there instead of getting it from the bid sorter

crates/relay/src/auctioneer/types.rs
```diff
@@ -40,6 +40,7 @@ pub struct GetPayloadResultData {
     pub to_publish: VersionedSignedProposal,
     pub trace: GetPayloadTrace,
     pub fork: ForkName,
+    pub builder_pubkey: Option<BlsPublicKeyBytes>,
```
ltitanb: why optional?

crates/relay/src/api/proposer/get_payload.rs
```diff
@@ -283,6 +298,39 @@ impl<A: Api> ProposerApi<A> {
...
+                                        reason,
+                                    )
+                                    .await
+                                {
+                                    error!(
+                                        %builder_pubkey,
+                                        %err,
+                                        %block_hash_clone,
+                                        "failed to demote builder in database! Pausing all optimistic submissions"
+                                    );
```
ltitanb: here you are not really pausing submission right? this is a copy of the demotion logic in `handle_simulation_result` however id rather we keep the demotion in one place. Let's make a new event , "BuilderDemotion" to be sent to the auctioneer , and refactor the existing logic to be used by both that event and the `handle_simulation_result`

ltitanb (changes requested):

crates/relay/src/auctioneer/bid_sorter.rs
tosynthegeek: > the builder pubkey should be already in the "local" payload we have in the payload entry, if not let's add it there instead of getting it from the bid sorter

Okay I will add `builder_pubkey` to `PayloadEntry` and pass it through to `handle_get_payload`, but since `new_gossip` payloads may not include a builder key, I may have to keep it as optional

crates/relay/src/api/proposer/get_payload.rs
tosynthegeek: > here you are not really pausing submission right? this is a copy of the demotion logic in `handle_simulation_result` however id rather we keep the demotion in one place. Let's make a new event , "BuilderDemotion" to be sent to the auctioneer , and refactor the existing logic to be used by both that event and the `handle_simulation_result`

Make sense, I will resume this once I get clarification from the first review

crates/relay/src/auctioneer/bid_sorter.rs
tosynthegeek: > > the builder pubkey should be already in the "local" payload we have in the payload entry, if not let's add it there instead of getting it from the bid sorter
> 
> Okay I will add `builder_pubkey` to `PayloadEntry` and pass it through to `handle_get_payload`, but since `new_gossip` payloads may not include a builder key, I may have to keep it as optional

@ltitanb please let me know

ltitanb: yes sounds good

ltitanb:
hey @tosynthegeek , yes apologies for the delay! If you could please resolve the branch conflicts then me and @Owen66 will review!

PR #316, opened 2025-10-24
moves block merging to the auctioneer

0w3n-d:


crates/relay/src/auctioneer/worker.rs
```diff
@@ -170,69 +150,23 @@ impl SubWorker {
...
-                        } else {
-                            let submission_data = SubmissionData {
-                                submission,
-                                version,
-                                merging_preferences,
-                                withdrawals_root,
-                                trace,
-                            };
+                        let submission_data = SubmissionData {
+                            submission: submission.clone(),
```
ltitanb:
```suggestion
                            submission,
```

crates/simulator/src/block_merging/mod.rs
```diff
@@ -62,9 +62,10 @@ impl BlockMergingApi {
         &self,
         request: BlockMergeRequestV1,
     ) -> Result<BlockMergeResponseV1, BlockMergingApiError> {
+        let based_block_hash = request.execution_payload.payload_inner.payload_inner.parent_hash;
```
ltitanb: how come we are taking the parent hash here? from how it is used elsewhere i thought the `base_block_hash` was the hash of the block that was merged?

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -0,0 +1,365 @@
...
+
+const MERGE_REQUEST_INTERVAL_MS: u64 = 50;
+
+pub struct BlockMerger {
+    curr_bid_slot: u64,
+    chain_info: ChainInfo,
+    local_cache: LocalCache,
+    best_merged_block: Option<BestMergedBlock>,
+    best_mergeable_orders: BestMergeableOrders,
+    base_blocks: HashMap<B256, BaseBlockData>,
```
ltitanb: worth adding a doc for what the map key is

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -0,0 +1,365 @@
...
+            } else {
+                None
+            }
+        })
+    }
+
+    pub fn update(&mut self, block: &ValidatedData<'_>) {
+        if let Some(merging_data) = &block.merging_data {
+            let mergeable_orders =
+                get_mergeable_orders(&block.submission, merging_data.clone()).ok();
```
ltitanb: these functions should be moved here now as they are not really used anywhere else. On a separate note, would think this can take quite a bit of time to do and it would block we should ideally block the event loop for as little as possible, so would keep it in the worker. One point would be whether it's worth delaying the submission reaching the bid sorter to wait for getting the mergeable orders. An alternative would for the worker to split up the submission data and mergeable orders

ltitanb: ideally we also dont silently error if theres an issue with getting the mergeable orders

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -0,0 +1,365 @@
...
+            }
+        }
+    }
+
+    pub fn should_request_merge(&self) -> bool {
+        let has_new_data = self.best_mergeable_orders.has_new_orders() ||
+            (self.best_mergeable_orders.has_orders() && self.has_new_base_block);
+        let time_since_last_request_ms =
+            utcnow_ms().saturating_sub(self.last_merge_request_time_ms);
+        has_new_data && time_since_last_request_ms >= MERGE_REQUEST_INTERVAL_MS
```
ltitanb: probably already optimzed by the compiler but think we can short circuit if `has_new_data` is false we early return and don't even take the timestamp

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -0,0 +1,365 @@
...
+    pub fn insert_orders(
+        &mut self,
+        slot: u64,
+        max_blobs_per_block: usize,
+        bid_value: U256,
+        mergeable_orders: MergeableOrders,
+    ) {
+        // If the orders are for a newer slot, reset current state.
+        if self.current_slot < slot {
+            self.reset(slot, max_blobs_per_block);
```
ltitanb: do we still need this ? since it's already called in `BlockMerger::on_new_slot`

crates/relay/src/auctioneer/get_header.rs
```diff
@@ -29,11 +29,61 @@ impl Context {
...
+            if merged_bid_higher(&merged_bid, &original_bid, time, *merging_preferences) {
+                return Ok(merged_bid);
+            }
+        }
+
+        Ok(original_bid)
     }
 }
+
+fn merged_bid_higher(
```
ltitanb: in `block_merger.get_header` there's a clone so we could move these checks in there? that way if the checks fail we avoid the clone

crates/relay/src/auctioneer/get_header.rs
```diff
@@ -29,11 +29,61 @@ impl Context {
...
         };
 
-        let Some(data) = entry.to_header_data() else {
+        let Some(original_bid) = entry.to_header_data() else {
             error!("failed to get header data from payload entry, this should never happen!");
             return Err(ProposerApiError::NoBidPrepared);
         };
 
-        Ok(data)
+        if self._config.block_merging_config.is_enabled {
```
ltitanb: nit `_config` was unused before so had the leading underscore, we can rename it `config` now that it is used

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    pub fn fetch_best_mergeable_orders(
+        &mut self,
+    ) -> Option<(&[MergeableOrderWithOrigin], &HashMap<B256, BlobWithMetadata>)> {
+        self.has_new_base_block = false;
+        self.last_merge_request_time_ms = utcnow_ms();
+        self.best_mergeable_orders.load()
+    }
+
+    pub fn prepare_merged_payload_for_storage(
```
ltitanb, line 146: this function also feels quite heavy, any way we can make it faster and avoid some of the cloning to avoid blocking the event loop?

crates/relay/src/auctioneer/get_header.rs
0w3n-d: done

crates/relay/src/auctioneer/get_header.rs
0w3n-d: done

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: nope. removed

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: done

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: replaced with a type for clarity

crates/simulator/src/block_merging/mod.rs
0w3n-d: that's a mistake. should have been block_hash. fixed

crates/relay/src/auctioneer/worker.rs
0w3n-d: done

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: OK I've moved it back to the worker for now

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: Removed a bit of the cloning but it's still cloning blobs. Can't avoid that but maybe could change how the payload is handled and lazily do this work later when get payload is called?

crates/simulator/src/block_merging/mod.rs
```diff
@@ -62,9 +62,10 @@ impl BlockMergingApi {
         &self,
         request: BlockMergeRequestV1,
     ) -> Result<BlockMergeResponseV1, BlockMergingApiError> {
+        let based_block_hash = request.execution_payload.payload_inner.payload_inner.block_hash;
```
ltitanb: nit for consistency with below
```suggestion
        let base_block_hash = request.execution_payload.payload_inner.payload_inner.block_hash;
```

crates/simulator/src/block_merging/mod.rs
```diff
@@ -62,9 +62,10 @@ impl BlockMergingApi {
...
         info!(
             target: "rpc::relay::block_merging",
-            block_hash=%request.execution_payload.payload_inner.payload_inner.block_hash,
+            block_hash=%based_block_hash,
```
ltitanb:
```suggestion
            %base_block_hash,
```

crates/simulator/src/block_merging/mod.rs
```diff
@@ -80,6 +81,7 @@ impl BlockMergingApi {
         // async fn to not be Send, which is required for spawning it.
         let (response, blob_versioned_hashes, request_cache) = self
             .merge_block(
+                based_block_hash,
```
ltitanb:
```suggestion
                base_block_hash,
```

crates/relay/src/auctioneer/worker.rs
```diff
@@ -170,69 +151,40 @@ impl SubWorker {
...
-                                            warn!(%err, "failed to get mergeable orders");
-                                            return;
-                                        }
-                                    };
-
-                                if mergeable_orders.orders.is_empty() {
-                                    return;
+                        let merging_data = merging_data.and_then(|data| {
+                            if let Submission::Full(ref signed_bid_submission) = submission {
+                                match get_mergeable_orders(signed_bid_submission, &data) {
```
ltitanb, line 159: this adds a considerable delay to the submission, let's add a TODO for a future PR to "split up" sending the payload to the auctioneer (so top bid update is still fast), and later send the mergeable orders which are less latency sensitive

crates/relay/src/auctioneer/context.rs
```diff
@@ -177,6 +184,26 @@ impl Context {
...
+    pub fn handle_merge_response(&mut self, response: BlockMergeResponse) {
+        let block_hash = response.execution_payload.block_hash;
+        let Some(original_payload) = self.payloads.get(&response.base_block_hash) else {
+            warn!(%block_hash, "could not fetch original payload for merged block");
+            return;
+        };
+
+        let original_payload_and_blobs = original_payload.payload_and_blobs.clone();
+
+        let Some((block_hash, payload)) = self
```
ltitanb: nit, do we need to return the block_hash here? should be the same as 
```rust
let block_hash = response.execution_payload.block_hash;
```

crates/relay/src/auctioneer/block_merger.rs
ltitanb: that seems a bit risky maybe, let's see how much overhead it actually adds

0w3n-d: added a TODO which will be one of the next things I pick up after I've added more metrics and tracing for block merging and done some tests

crates/simulator/src/block_merging/mod.rs
0w3n-d: done

crates/simulator/src/block_merging/mod.rs
0w3n-d: done

crates/relay/src/auctioneer/worker.rs
0w3n-d: done

crates/relay/src/auctioneer/context.rs
0w3n-d: fixed

PR #321, opened 2025-11-03
adds submission type header

0w3n-d:


crates/relay/src/auctioneer/worker.rs
```diff
@@ -312,14 +314,24 @@ impl SubWorker {
...
+                let sub_with_merging: SignedBidSubmissionWithMergingData = decoder.decode(body)?;
+                let upgraded = sub_with_merging.maybe_upgrade_to_fulu(self.chain_info.current_fork_name());
+                (upgraded.submission, Some(upgraded.merging_data))
             } else {
-                let payload: SignedBidSubmission = decoder.decode(body)?;
-                let payload = payload.maybe_upgrade_to_fulu(self.chain_info.current_fork_name());
-                (payload, None)
+                let sub: SignedBidSubmission = decoder.decode(body)?;
+                let upgraded = sub.maybe_upgrade_to_fulu(self.chain_info.current_fork_name());
+                let merging_data = if has_mergeable_data {
```
ltitanb: this is effectively a new behaviour so should we not have a specific `SubmissionType` ? as ideally we start deprecating all these headers and move to the single `SubmissionType`

0w3n-d: OK makes sense. I've added that and refactored a bit.

crates/relay/src/auctioneer/decoder.rs
```diff
@@ -17,13 +17,33 @@ use http::{
...
 
-use crate::api::builder::{api::MAX_PAYLOAD_LENGTH, error::BuilderApiError};
+use crate::api::{
+    HEADER_SUBMISSION_TYPE,
+    builder::{api::MAX_PAYLOAD_LENGTH, error::BuilderApiError},
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumString, AsRefStr)]
+#[strum(serialize_all = "lowercase")]
+pub enum SubmissionType {
```
gd-0, line 34: imo submission type is a bit general. can we changing this to merging type?

0w3n-d: It needs to be general as it covers all types including `DehydratedBidSubmission`. The idea is to deprecate `x-hydrate` and `x-mergeable` and just use this one header to tell the handler what to expect. It's also extensible to cover any future changes to the api.

gd-0 (approved):

PR #322, opened 2025-11-04
fix: prevent DB deadlock on bid save

0w3n-d:
Remove ON CONFLICT, which requires a lock and can cause data loss if a deadlock occurs. Add a retry mechanism to further reduce the chance of errors leading to lost data.

Remove the uniqueness constraint from block submissions to speed up inserts and reduce lock contention. This removes built-in deduplication, but we already perform a dedupe when a bid is received. Some duplicates can still slip through, so add an additional dedupe step in the data API before returning results to the user.

Remove get_header_trace and submission_trace.
get_header times can be flattened into the get_header table, and submission_trace is obsolete since its recorded times have been zero for a while.

### DIFF
--- a/crates/relay/src/api/relay_data/api.rs
+++ b/crates/relay/src/api/relay_data/api.rs
@@ -92,8 +92,10 @@ impl DataApi {
             .await
         {
             Ok(result) => {
+                let mut seen = std::collections::HashSet::with_capacity(result.len());
                 let response = result
                     .into_iter()
+                    .filter(|b| seen.insert(b.bid_trace.block_hash))
                     .map(|b| b.into())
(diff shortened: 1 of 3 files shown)


crates/relay/src/api/relay_data/api.rs
```diff
@@ -92,8 +92,10 @@ impl DataApi {
             .await
         {
             Ok(result) => {
+                let mut seen = std::collections::HashSet::with_capacity(result.len());
```
vladimir-ea, line 95: what happens to the `seen` set? cannot see it used anywhere?

vladimir-ea: oh i see - its de-duplicating

PR #324, opened 2025-11-19
flattens delivered payloads

0w3n-d:
Currently the delivered payloads query is very slow due to the join on the block submission table. This change removes the need for that join by adding the missing data to the delivered payloads table.

### DIFF
--- a/crates/common/src/local_cache.rs
+++ b/crates/common/src/local_cache.rs
@@ -105,7 +105,7 @@ impl LocalCache {
         let primev_proposers = Arc::new(DashSet::with_capacity(MAX_PRIMEV_PROPOSERS));
         let kill_switch = Arc::new(AtomicBool::new(false));
         let proposer_duties = Arc::new(RwLock::new(Vec::new()));
-        let merged_blocks = Arc::new(DashMap::with_capacity(10000));
+        let merged_blocks = Arc::new(DashMap::with_capacity(1000));
 
         Self {
             inclusion_list: Default::default(),
--- a/crates/relay/src/auctioneer/get_payload.rs
+++ b/crates/relay/src/auctioneer/get_payload.rs
@@ -38,7 +38,7 @@ impl Context {
 
         let block_hash = payload.execution_payload.execution_payload.block_hash;
         let entry = PayloadEntry::new_gossip(payload);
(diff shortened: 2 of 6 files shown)


crates/common/src/local_cache.rs
```diff
@@ -105,7 +105,7 @@ impl LocalCache {
         let primev_proposers = Arc::new(DashSet::with_capacity(MAX_PRIMEV_PROPOSERS));
         let kill_switch = Arc::new(AtomicBool::new(false));
         let proposer_duties = Arc::new(RwLock::new(Vec::new()));
-        let merged_blocks = Arc::new(DashMap::with_capacity(100));
+        let merged_blocks = Arc::new(DashMap::with_capacity(10000));
```
gd-0: seems overkill here no? when would we build 10k merged blocks?

crates/relay/src/auctioneer/get_payload.rs
```diff
@@ -37,15 +38,6 @@ impl Context {
...
         let block_hash = payload.execution_payload.execution_payload.block_hash;
         let entry = PayloadEntry::new_gossip(payload);
-
-        if let Some(curr) = self.payloads.get(&block_hash) &&
-            curr.bid_data.is_some()
-        {
-            // we received this block both locally and via gossip, keep the local one so we can
-            // serve get_header on it
-            return;
-        }
```
gd-0, line 47: what;s the reasoning for removing this? are we not adding a dup here for no reason?

crates/common/src/local_cache.rs
0w3n-d: Yeah don't know what I was thinking there. Reduced it.

crates/relay/src/auctioneer/get_payload.rs
0w3n-d: I removed it because the previous logic was based on gossiped payloads having no bid_data, but I've added the missing bid_data to the gossip so that's not longer needed.

However, I take your point about the needless extra work here. I've changed how the insert works to avoid that.

0w3n-d:
Going to merge this as it's need to address a production issue. If there's anything else that need changing I can fix in a follow up PR.

PR #327, opened 2025-11-28
add inclusion list txs to merged blocks

0w3n-d:


crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -92,6 +93,24 @@ impl BlockMerger {
         self.last_merge_request_time_ms = 0;
     }
 
+    pub fn add_inclusion_list(
+        &mut self,
+        inclusion_list: &InclusionListWithMetadata,
+    ) {
+        inclusion_list.txs.iter().for_each(|o| {
+            match self.best_mergeable_orders.order_map.entry(o.bytes.clone().into()) {
```
gd-0: what does this .into do? Is it going from Bytes to vec? if so that allocs.

gd-0: also we can just use get() rather than entry api to remove the need for ownership on the bup check.
When testing entry it's actually always worse or best case equal than doing a get() and then insert if missing

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -92,6 +93,24 @@ impl BlockMerger {
...
+                Entry::Occupied(_) => {}
+                Entry::Vacant(e) => {
+                    self.best_mergeable_orders.has_new_orders = true;
+                    let index = self.best_mergeable_orders.best_orders.len();
+                    // TODO: Maybe use relay address as origin?
+                    self.best_mergeable_orders.best_orders.push(MergeableOrderWithOrigin::new(Address::default(), e.key().clone()));
+                    e.insert(OrderMetadata { value: U256::ZERO, index });
```
gd-0: yep lets definitely set relay address as that address already gets paid. So no extra cost to transfer

crates/relay/src/auctioneer/mod.rs
```diff
@@ -550,6 +550,10 @@ impl State {
             payload_attributes_map.insert(update.parent_hash, update);
         }
 
+        if let Some(il) = &il {
+            ctx.block_merger.add_inclusion_list(il);
+        }
```
gd-0, line 530: lets put behind a flag?

gd-0: if we want to properly enforce the IL we also would require the mergeable block to win here. Think this PR is fine as a V1 but to make this fully productionised we need to think through this more

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: I noticed the titanrelay.eth ENS address that's used for collateral and the relay pubkey / signing key don't match so it's hard to say which one is the relay address. So I just address a relay address field to the IL config so it's flexible.

crates/relay/src/auctioneer/mod.rs
0w3n-d: I put the check inside add_inclusion_list. Also, the IL stuff in general is optional in the config so it's either all on or all off.
But the check to enfore the IL on the builder doesn't seem to be implemented yet so I can turn on the IL stuff for now in the hoodi relay to test the block merge stuff and we can discuss later how it should all work in production.

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: The into just wraps the bytes in an enum so no alloc

I've replayed the match on the entry with entry.or_insert _with which seems to be optimal here

PR #333, opened 2025-12-03
Add payloads for bid submissions with adjustments

ninaiad:


crates/types/src/bid_adjustment_data.rs
```diff
@@ -0,0 +1,224 @@
...
+#[serde(untagged)]
+pub enum BidAdjustmentData {
+    Original(BidAdjData),
+    GasUsed(BidAdjDataWithGasUsed),
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Encode, Decode, Clone)]
+#[ssz(enum_behaviour = "transparent")]
+#[serde(untagged)]
+pub enum BidAdjustmentDataV2 {
```
ninaiad: I copied these (V2) from the builder, but they aren't used here yet, unsure if necessary

crates/types/src/bid_adjustment_data.rs
```diff
@@ -0,0 +1,224 @@
...
+    Original(BidAdjDataV2),
+    GasUsed(BidAdjDataV2WithGasUsed),
+}
+
+/// Adjustment data compatible with Ultrasound spec.
+/// a subset of the now bigger DecodedBidAdjData ( which now includes extra
+/// payment as well )
```
vladimir-ea: I think this comment should be removed - it references bid-submitter code I think

crates/types/src/hydration.rs
```diff
@@ -318,6 +343,99 @@ impl DehydratedBidSubmissionFulu {
...
+
+        let submission = SignedBidSubmission::Fulu(SignedBidSubmissionFulu {
+            message: self.message,
+            execution_payload: Arc::new(self.execution_payload),
+            blobs_bundle: Arc::new(BlobsBundle::V2(sidecar)),
+            execution_requests: self.execution_requests,
+            signature: self.signature,
+        });
+
+        Ok(HydratedData { submission, tx_cache_hits, blob_cache_hits, tx_root: self.tx_root })
```
vladimir-ea, line 379: does this duplicate a function in `DehydratedBidSubmission`? It would be preferable not to duplicate the function so we don't have to remember to make future changes in 2 places.

Can we share the logic somehow? Perhaps define a trait that both structs implement? e.g. trait with method to access payload, blobs_bundle etc.?

crates/types/src/bid_submission.rs
```diff
@@ -225,12 +225,23 @@ impl Decode for SignedBidSubmissionFulu {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode)]
+pub struct SignedBidSubmissionFuluWithAdjustments {
```
0w3n-d: I think we should implement this like we do [here](https://github.com/gattaca-com/helix/blob/develop/crates/types/src/block_merging.rs#L17) for the block merging data, if we're just receiving some additional data along side the SignedBidSubmission.
This would avoid the need to add an additional match to each of the SignedBidSubmission methods.

ninaiad: It looks like we can't do that, because the payload with adjustments looks exactly like the one without it, but has an additional field (see testdata) and we cannot ssz decode it from that into a nested struct like the one with merging data

0w3n-d: Maybe we should replace `SignedBidSubmissionWithMergingData` with something more general. Not sure what to call it but it would have the bid submission field and then two optional fields. One for merge data and one for bid adj data and we can extent it with other optional fields in the future if other data needs to be sent?

We could then have the same but for dehydrated stuff? Or just have one struct and make the bid submission and enum?

ltitanb: i think after decoding we already have this which has all the data https://github.com/gattaca-com/helix/blob/develop/crates/relay/src/auctioneer/types.rs#L46-L52
just a matter of cleaninig up the decoding part in the worker

PR #335, opened 2025-12-08
Add api for applying bid adjustments

ninaiiad:


crates/relay/src/auctioneer/bid_adjustor.rs
```diff
@@ -0,0 +1,24 @@
...
+    ) -> Option<PayloadHeaderData>;
+}
+
+#[derive(Clone)]
+pub struct DefaultBidAdjustor;
+
+impl BidAdjustor for DefaultBidAdjustor {
+    fn try_apply_adjustments(
+        &self,
+        _bid: &PayloadHeaderData,
```
vladimir-ea: I think we need the full block here, not just the header since we need to modify the transactions in the block and then create and sign a new block.

crates/relay/src/auctioneer/get_header.rs
```diff
@@ -34,10 +34,16 @@ impl Context {
...
+        let Some(merged_bid) = self.block_merger.get_header(&original_bid) else {
+            return Ok(original_bid);
+        };
+
+        let Some(adjusted_bid) =
+            self.bid_adjustor.try_apply_adjustments(&merged_bid, &entry.bid_adjustment_data)
+        else {
             return Ok(merged_bid);
-        }
+        };
```
vladimir-ea, line 34: not sure this is correct - it looks like we would only try to adjust `merged_bid`s but we should try to adjust blocks that cannot be merged also.

vladimir-ea: probably if there is a merged block we use that, then if there is none we should try to adjust - I am not sure we would adjust merged blocks?

crates/relay/src/auctioneer/bid_adjustor.rs
vladimir-ea: oh, my bad, I see that this `PayloadHeaderData` does contain the payload.

crates/relay/src/auctioneer/types.rs
```diff
@@ -129,23 +131,34 @@ impl PayloadEntry {
...
+            bid_adjustment_data,
         }
     }
 
     pub fn new_gossip(data: BroadcastPayloadParams) -> Self {
-        Self { payload_and_blobs: data.execution_payload, bid_data: data.bid_data }
+        Self {
+            payload_and_blobs: data.execution_payload,
+            bid_data: data.bid_data,
+            bid_adjustment_data: None, // the gossiped bid should already be adjusted (?)
```
vladimir-ea: I think that is correct. @0w3n-d ?

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -148,6 +149,7 @@ impl BlockMerger {
         response: BlockMergeResponse,
         original_payload: Arc<PayloadAndBlobs>,
         builder_pubkey: BlsPublicKeyBytes,
+        bid_adjustment_data: Option<BidAdjustmentData>,
```
0w3n-d: Should remove this as merged blocks won't be adjusted

crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -216,14 +218,15 @@ impl BlockMerger {
         let new_bid = PayloadHeaderData {
             payload_and_blobs: payload_and_blobs.clone(),
             bid_data: bid_data.clone(),
+            bid_adjustment_data: bid_adjustment_data.clone(),
```
0w3n-d: and remove this

crates/relay/src/auctioneer/get_header.rs
```diff
@@ -36,8 +36,13 @@ impl Context {
 
         if let Some(merged_bid) = self.block_merger.get_header(&original_bid) {
             return Ok(merged_bid);
-        }
+        };
+
+        let adjusted_bid = self
+            .bid_adjustor
+            .try_apply_adjustments(&original_bid, &entry.bid_adjustment_data)
```
0w3n-d: should return the adjusted PayloadEnty here and store it in self.payloads so if the header wins and get payload is called we have the the adjusted block to publish

crates/relay/src/auctioneer/types.rs
0w3n-d: Yeah when the BroadcastPayloadParams are received by the other relay they'll only be read if get payload is callled. They won't be read in get header as this bid won't have been included in the bid sorter, so there's no need to send them here.

PR #337, opened 2025-12-10
feat(data-api): Add proposer_header_delivered endpoint

ckoopmann:
Many analysts as well as websites such as [relayscan](https://relayscan.io/overview?t=1h) use payload deliveries to determine market share of relays. 
This is misleading since a lot of these payload deliveries are from relays that did not deliver the header. 
To determine the share of slots in which a given relay was selected by the proposer in the header "auction" it'd be more interesting to look at headers delivered.

To enable this, this PR adds a "proposer_headers_delivered" endpoint that returns all headers (block_hash, value) delivered for a given slot auction. 

Note that ideally these results would be filtered to contain only headers delivered to the actual proposer, however this would require adding a mechanism to identify which header calls come from the proposer. So I left it out for this initial implementation.

### DIFF
--- a/crates/relay/src/database/postgres/postgres_db_service.rs
+++ b/crates/relay/src/database/postgres/postgres_db_service.rs
@@ -2260,8 +2260,6 @@ impl PostgresDatabaseService {
         Ok(())
     }
 
-    // TODO: proposer_send_timestamp_ms is not tracked. To support this, the get_header table
-    // would need to store the timestamp from the proposer's request header.
     #[instrument(skip_all)]
     pub async fn get_proposer_header_delivered(
         &self,
@@ -2278,7 +2276,7 @@ impl PostgresDatabaseService {
                 proposer_pubkey,
                 value
             FROM get_header
-            WHERE 1 = 1
+            WHERE mev_boost = true
         ",
         );
 


gd-0 (approved):
nice - thanks for the PR!

looks good to me. I haven't touched the data API much though so would wait for @0w3n-d's review

crates/relay/src/database/postgres/postgres_db_service.rs
```diff
@@ -2176,4 +2178,86 @@ impl PostgresDatabaseService {
...
+        let mut query = String::from(
+            "
+            SELECT
+                slot_number,
+                parent_hash,
+                block_hash,
+                proposer_pubkey,
+                value
+            FROM get_header
+            WHERE 1 = 1
```
0w3n-d: We do have a way to detect if the header call came from the proposer. It may have a couple of false positives but I think it's worth adding `mev_boost = true` here to avoid returning all the records that come from python user agents and so on.

crates/relay/src/database/postgres/postgres_db_service.rs
```diff
@@ -2176,4 +2178,86 @@ impl PostgresDatabaseService {
...
+
+        let params_refs: Vec<&(dyn ToSql + Sync)> =
+            db_params.iter().map(|p| &**p as &(dyn ToSql + Sync)).collect();
+
+        let rows = self.pool.get().await?.query(&query, &params_refs[..]).await?;
+
+        let results = rows
+            .iter()
+            .map(|row| ProposerHeaderDeliveredResponse {
+                slot: row.get::<_, Option<i32>>("slot_number").map(|s| s.to_string()),
```
0w3n-d: We would normally use the actual types in a response struct like ProposerHeaderDeliveredResponse and let serde handle converting the type to json at the application boundary. It avoids the to_string() allocation.

crates/relay/src/database/postgres/postgres_db_service.rs
ckoopmann: Great point. Thanks.
Implemented here:  https://github.com/gattaca-com/helix/pull/337/changes/9bd2781ba45ec4ad4b84268d561d99cfcca6c7fc

crates/relay/src/database/postgres/postgres_db_service.rs
ckoopmann: Thanks, good point.
I removed the serialization here and instead rely on serde: https://github.com/gattaca-com/helix/pull/337/commits/ab0e35979527f9bf7e116599192ad6f5452c7864

ckoopmann: Note that I still kept the `Option` types, since none of columns in the get_header table seem to have a "NOT NULL" constraint.
It seems elsewhere in such cases you rely on the app never adding such rows into the database even where the schema would allow it. 
Do let me know if you want me to change the code here to align with those practices.
('value' needs to be kept optional though since it will be Null for records inserted into get_header before deploying this PR)

crates/relay/src/database/postgres/postgres_db_service.rs
ckoopmann: @0w3n-d Just to make sure: "mev_boost" will be true for all validators including those running commit-boost or Vouch. Correct ?

PR #345, opened 2025-12-16
TopbidV2 with Auction open message 

louisponet:
This allows the relay to communicate when the getheader call is made and the auction is thus open.

### DIFF
--- a/crates/relay/src/api/builder/top_bid.rs
+++ b/crates/relay/src/api/builder/top_bid.rs
@@ -81,14 +81,10 @@ async fn socket_loop_body(socket: &mut WebSocket, interval: &mut time::Interval)
                         return true;
                     }
                 },
-                Some(Ok(Message::Pong(_))) => {
+                Some(Ok(Message::Pong(_))) | Some(Ok(Message::Binary(_))) | Some(Ok(Message::Text(_))) => {
                 },
                 Some(Ok(Message::Close(_))) => {
                     return true;
-                },
-                Some(Ok(Message::Binary(_))) => {
-                },
-                Some(Ok(Message::Text(_))) => {
                 },
                 Some(Err(e)) => {
                     error!("Error in WebSocket connection: {}", e);


crates/relay/src/api/builder/top_bid.rs
```diff
@@ -34,6 +38,71 @@ impl<A: Api> BuilderApi<A> {
...
+                },
+                Some(Ok(Message::Close(_))) => {
+                    debug!("Received close frame.");
+                    return true;
+                },
+                Some(Ok(Message::Binary(_))) => {
+                    debug!("Received Binary frame.");
+                },
+                Some(Ok(Message::Text(_))) => {
+                    debug!("Received Text frame.");
```
gd-0: assuming copy pastad but these debugs are completely useless. we should remove

crates/common/src/api_provider.rs
```diff
@@ -39,3 +40,8 @@ impl ApiProvider for DefaultApiProvider {
         Ok(TimingResult { sleep_time: None, is_mev_boost: false })
     }
 }
+
+#[derive(Clone, Copy, Decode, Encode)]
+pub struct GetHeaderInfo {
```
ltitanb, line 45: is there anything we could add here? for example a timestamp

ltitanb: also why would we ever send this with `false`?

crates/relay/src/api/builder/top_bid.rs
```diff
@@ -50,58 +119,95 @@ impl<A: Api> BuilderApi<A> {
...
+            Ok(bid) = bid_stream.recv() => {
+                if socket.send(Message::Binary(TopBidV2::TopBid(bid).as_ssz_bytes_fast().into())).await.is_err() {
+                    error!("Failed to send bid. Disconnecting.");
+                    break;
+                }
+            },
+            Ok(getheader) = getheader_call_stream.recv() => {
+                if socket.send(Message::Binary(TopBidV2::AuctionOpen(getheader.called).as_ssz_bytes_fast().into())).await.is_err() {
+                    error!("Failed to send bid. Disconnecting.");
+                }
```
gd-0, line 205: should break here as well right? means the conn is closed

crates/types/benches/topbid_ssz.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub block_number: u64,
+    pub block_hash: B256,
+    pub parent_hash: B256,
+    pub builder_pubkey: BlsPublicKeyBytes,
+    pub fee_recipient: Address,
+    pub value: U256,
+}
+
+impl TopBidUpdate {
+    const SSZ_SIZE: usize = 188;
```
gd-0: can't you use `size_of::<TopBidUpdate>` here?

gd-0: wait - why is this defined in the bench? and not using the actual top bid update??

crates/relay/src/api/builder/top_bid.rs
```diff
@@ -50,58 +119,95 @@ impl<A: Api> BuilderApi<A> {
...
+#[ssz(enum_behaviour = "union")]
+pub enum TopBidV2 {
+    TopBid(TopBidUpdate),
+    AuctionOpen(bool),
+}
+
+impl TopBidV2 {
+    pub fn as_ssz_bytes_fast(&self) -> Vec<u8> {
+        const SIZE_TOPBID: usize = 189;
+        const SIZE_AUCTION_OPEN: usize = 2;
```
gd-0, line 151: should use size_of the 2 types + 1 for the enum size

crates/relay/src/api/builder/top_bid.rs
```diff
@@ -50,58 +119,95 @@ impl<A: Api> BuilderApi<A> {
...
+///
+/// This function operates in an asynchronous loop until the WebSocket connection is closed either
+/// due to an error or when the auction ends. It returns after the socket has been closed, logging
+/// the closure status.
+async fn push_top_bids_v2(
+    mut socket: WebSocket,
+    mut bid_stream: tokio::sync::broadcast::Receiver<TopBidUpdate>,
+    mut getheader_call_stream: tokio::sync::broadcast::Receiver<GetHeaderInfo>,
+) {
+    let _conn = TopBidMetrics::connection();
```
ltitanb: would add a new metric here

crates/common/src/api_provider.rs
louisponet: Thought that yeah, but it's atm really just an event. Not sure what the usecase for a timestamp would be.

crates/relay/src/api/builder/top_bid.rs
louisponet: Doesn't work like that. SizeOf for topbid is 192

crates/types/benches/topbid_ssz.rs
louisponet: Because Topbid isn't in types. Benches are, can just remove bench tbh

crates/relay/src/api/builder/top_bid.rs
louisponet: Okay

crates/types/benches/topbid_ssz.rs
louisponet: removed the bench, useless anyway

gd-0 (approved):

PR #350, opened 2025-12-31
adjusted blocks simulation

ninaiiad:
changes in PR:
- PayloadEntry refactoring to add some fields necessary for simulation request
- sending adjusted blocks for simulation
- if simulation fails, adjustments are disabled in db (can only enable manually for now, but the running code will pick up the new value because it's querying db every 5s)
    - if db is unavailable adjustments are disabled via an atomic flag (to enable we'll need to redeploy the instance)

### DIFF
--- a/crates/relay/src/auctioneer/context.rs
+++ b/crates/relay/src/auctioneer/context.rs
@@ -58,8 +58,8 @@ pub struct Context<B: BidAdjustor> {
 const EXPECTED_PAYLOADS_PER_SLOT: usize = 5000;
 const EXPECTED_BUILDERS_PER_SLOT: usize = 200;
 
-const DB_CHECK_INTERVAL_SEC: u64 = 1;
-const ADJUSTMENTS_DRY_RUN_INTERVAL_MS: u64 = 500;
+const DB_CHECK_INTERVAL: Duration = Duration::from_secs(1);
+const ADJUSTMENTS_DRY_RUN_INTERVAL: Duration = Duration::from_millis(500);
 
 impl<B: BidAdjustor> Context<B> {
     pub fn new(
@@ -279,7 +279,7 @@ impl<B: BidAdjustor> Context<B> {
         failsafe_triggered: Arc<AtomicBool>,
     ) {
         spawn_tracked!(async move {
-            let mut interval = tokio::time::interval(Duration::from_secs(DB_CHECK_INTERVAL_SEC));
+            let mut interval = tokio::time::interval(DB_CHECK_INTERVAL);
             loop {
                 interval.tick().await;
 
@@ -310,8 +310,7 @@ impl<B: BidAdjustor> Context<B> {
         adjustments_enabled: Arc<AtomicBool>,
     ) {
         spawn_tracked!(async move {
-            let mut interval =
-                tokio::time::interval(Duration::from_millis(ADJUSTMENTS_DRY_RUN_INTERVAL_MS));
+            let mut interval = tokio::time::interval(ADJUSTMENTS_DRY_RUN_INTERVAL);
             loop {
                 interval.tick().await;
 


crates/relay/src/auctioneer/context.rs
```diff
@@ -117,7 +132,18 @@ impl<B: BidAdjustor> Context<B> {
...
-            if self.cache.demote_builder(&builder) {
+            if self.payloads.get(&block_hash).is_some_and(|bid| bid.is_adjusted()) {
+                warn!(%builder, %block_hash, %err, "Block simulation resulted in an error. Disabling adjustments...");
+
+                let db = self.db.clone();
+                let adjustments_enabled = self.adjustments_enabled.clone();
+                spawn_tracked!(async move {
+                    if let Err(err) = db.disable_adjustments().await {
+                        error!(%block_hash, %err, "failed to disable adjustments in database, pulling the failsafe trigger");
+                        adjustments_enabled.store(false, Ordering::Relaxed);
```
gd-0: lets immediately switch adjustments_enabled to false. I don't see why we'd need to wait for a failure
as long as we make sure to renable if we read that the db state has changed from disabled->enabled.

gd-0: lets also add a metric for this similar to the demotion_count() ?

gd-0 (changes requested):
looking good. just some points around the atomic

crates/relay/src/auctioneer/context.rs
```diff
@@ -91,7 +95,18 @@ impl<B: BidAdjustor> Context<B> {
             block_merger,
         };
 
-        Self { chain_info, cache, unknown_builder_info, slot_context, db, config, bid_adjustor }
+        let adjustments_enabled = Arc::new(AtomicBool::new(true));
```
gd-0: this needs to be set to whatever the value is in the db at the time i think? otherwise a restart overwrites the value

crates/relay/src/database/postgres/postgres_db_service.rs
```diff
@@ -2341,4 +2341,37 @@ impl PostgresDatabaseService {
...
+        let mut client = self.high_priority_pool.get().await?;
+        let transaction = client.transaction().await?;
+
+        transaction.execute("UPDATE relay_info SET adjustments_enabled = FALSE", &[]).await?;
+
+        record.record_success();
+        Ok(())
+    }
+
+    pub async fn check_adjustments_enabled(&self) -> Result<bool, DatabaseError> {
```
gd-0, line 2360: is this ever used?? where do we fetch from this and set the atomic?

crates/relay/src/auctioneer/context.rs
```diff
@@ -117,7 +147,23 @@ impl<B: BidAdjustor> Context<B> {
...
+            if self.adjustments_enabled.load(Ordering::Relaxed) &&
+                self.payloads.get(&block_hash).is_some_and(|bid| bid.is_adjusted())
+            {
+                warn!(%builder, %block_hash, %err, "Block simulation resulted in an error. Disabling adjustments...");
+
+                SimulatorMetrics::disable_adjustments();
+                self.adjustments_enabled.store(false, Ordering::Relaxed);
+
+                let db = self.db.clone();
+                let failsafe_trigger = self.adjustments_failsafe_trigger.clone();
```
gd-0: what is the point of this failsafe? unclear to me why it's actually required?

crates/relay/src/auctioneer/context.rs
```diff
@@ -214,6 +260,44 @@ impl<B: BidAdjustor> Context<B> {
...
+        spawn_tracked!(async move {
+            let mut interval =
+                tokio::time::interval(Duration::from_secs(ADJUSTMENTS_DRY_RUN_INTERVAL_SEC));
+            loop {
+                interval.tick().await;
+
+                // TODO: does this need to run if adjustments are disabled?
+                if let Err(e) = auctioneer.try_send(Event::DryRunAdjustments) {
+                    error!("failed to send adjustments dry run request: {}", e);
+                }
```
gd-0: nope

crates/relay/src/auctioneer/context.rs
```diff
@@ -214,6 +260,44 @@ impl<B: BidAdjustor> Context<B> {
         };
         self.payloads.insert(block_hash, payload);
     }
+
+    fn spawn_check_flag_task(
```
gd-0, line 276: i think this probably makes more sense to be in the housekeeper task

crates/relay/src/auctioneer/mod.rs
```diff
@@ -338,6 +346,29 @@ impl State {
                 ctx.handle_simulation_result((id, None));
             }
 
+            // Dry run adjusting & validating bids
+            (_, Event::DryRunAdjustments) => {
+                let Some(best_block_hash) = ctx.bid_sorter.get_any_header() else {
+                    warn!("adjustments dry run - no bids present yet");
+                    return;
+                };
```
gd-0: we should do this on the top-bid at that time rather than any header

ninaiiad: it is a top bid, but for a random "fork" - the bid_sorter maintains multiple top bids pointing to different parent hashes and on the get_header call the parent hash is passed explicitly to choose the fork

crates/relay/src/auctioneer/context.rs
ninaiiad: it's the same thing we do for builder demotions currently - if a write to db fails a separate flag is set that takes priority over one which mirrors db and can get invalid on a reread
I'm not sure if it does get used much for demotions, if not, can remove this

crates/relay/src/auctioneer/context.rs
ninaiiad: I didn't think so because it looks like housekeeper is mostly about managing slots and validator info
I could move it there but it's lightweight enough and it affects internal data structures that seemed convenient to keep in the same file

crates/relay/src/auctioneer/context.rs
gd-0: ah right i see. reread would set it back to true. makes sense to keep it. 
Do we have any more critical alert than an error? e.g., some way to log to discord if this happens?

crates/relay/src/auctioneer/context.rs
gd-0: fair enough. happy to keep

crates/relay/src/auctioneer/mod.rs
gd-0: ah. lets specify that then

crates/relay/src/auctioneer/context.rs
ninaiiad: will add

crates/relay/src/auctioneer/types.rs
```diff
@@ -124,40 +140,74 @@ impl PayloadEntry {
...
-        Self {
-            payload_and_blobs: data.execution_payload,
-            bid_data: data.bid_data,
-            bid_adjustment_data: None,
-            submission_version: None,
+    pub fn new_gossip(payload_and_blobs: PayloadAndBlobs, bid_data: PayloadBidData) -> Self {
+        Self::Gossip(GossipedPayload { payload_and_blobs, bid_data })
+    }
+
+    pub fn is_adjusted(&self) -> bool {
```
ltitanb, line 160:
```suggestion
    pub fn is_adjusted(&self) -> bool {
        matches!(self, Self::Submission(s) if s.is_adjusted)
    }
```

crates/relay/src/auctioneer/types.rs
```diff
@@ -106,15 +108,29 @@ impl Submission {
     }
 }
 
-/// From a SignedBidSubmission, keep only the fields needed to serve get_header and get_payload,
-/// some fields are optional because payloads also arrive via gossip and we only gossip
-/// PayloadAndBlobs
 #[derive(Clone)]
-pub struct PayloadEntry {
-    pub payload_and_blobs: Arc<PayloadAndBlobs>,
+pub struct GossipedPayload {
```
ltitanb: nit for consistency
```suggestion
pub struct GossipPayload {
```

crates/relay/src/auctioneer/types.rs
```diff
@@ -124,40 +140,74 @@ impl PayloadEntry {
...
 
     pub fn value(&self) -> &U256 {
-        &self.bid_data.value
+        match &self {
+            Self::Submission(s) => s.signed_bid_submission.value(),
+            Self::Gossip(s) => &s.bid_data.value,
+        }
+    }
+
+    pub fn payload_and_blobs(&self) -> PayloadAndBlobs {
```
ltitanb, line 171: i would still maintain the `_ref` form so the caller can decide when to clone. Even if it's just a couple of atomics we sometimes don't need that eg in a few places we only reference the parent hash

crates/relay/src/auctioneer/types.rs
```diff
@@ -124,40 +140,74 @@ impl PayloadEntry {
...
 
     pub fn execution_payload(&self) -> &ExecutionPayload {
-        &self.payload_and_blobs.execution_payload
+        match self {
+            Self::Submission(bid) => bid.signed_bid_submission.execution_payload_ref(),
+            Self::Gossip(bid) => &bid.payload_and_blobs.execution_payload,
+        }
     }
 
     pub fn execution_payload_make_mut(&mut self) -> &mut ExecutionPayload {
```
ltitanb, line 205: seems this is never called? let's just delete the method if unused?

crates/relay/src/auctioneer/types.rs
```diff
@@ -168,16 +218,28 @@ impl PayloadEntry {
     pub fn into_builder_bid_slow(self) -> BuilderBid {
```
ltitanb, line 217: note that this method takes `self` not `&self` so all the clones should be avoidable

crates/relay/src/auctioneer/types.rs
```diff
@@ -168,16 +218,28 @@ impl PayloadEntry {
...
+            ),
+            Self::Gossip(s) => (
+                None,
+                None,
+                s.bid_data.execution_requests.clone(),
+                s.payload_and_blobs.blobs_bundle.commitments().clone(),
+            ),
+        };
+
+        let header = self.execution_payload().clone().to_header(withdrawals_root, tx_root);
```
ltitanb: `to_header` takes `&self` so no need to clone here
```suggestion
        let header = self.execution_payload().to_header(withdrawals_root, tx_root);
```

crates/relay/src/auctioneer/context.rs
```diff
@@ -214,6 +268,50 @@ impl<B: BidAdjustor> Context<B> {
...
```
ltitanb, line 276: we should log something whenever any of this flags flips in value

crates/relay/src/auctioneer/context.rs
```diff
@@ -214,6 +268,50 @@ impl<B: BidAdjustor> Context<B> {
...
+        auctioneer: crossbeam_channel::Sender<Event>,
+        adjustments_enabled: Arc<AtomicBool>,
+    ) {
+        spawn_tracked!(async move {
+            let mut interval =
+                tokio::time::interval(Duration::from_secs(ADJUSTMENTS_DRY_RUN_INTERVAL_SEC));
+            loop {
+                interval.tick().await;
+
+                if !adjustments_enabled.load(Ordering::Relaxed) {
```
gd-0, line 317: could check the failsafe here as well?

crates/relay/src/auctioneer/get_header.rs
```diff
@@ -33,11 +35,18 @@ impl<B: BidAdjustor> Context<B> {
             return Ok(merged_bid);
         };
 
-        if let Some(adjusted_bid) = self.bid_adjustor.try_apply_adjustments(original_bid) {
-            let block_hash = adjusted_bid.block_hash();
-            self.payloads.insert(*block_hash, adjusted_bid.clone());
+        if self.adjustments_enabled.load(Ordering::Relaxed) {
```
gd-0, line 43: think we need to check the failsafe here as well?
```suggestion
        if self.adjustments_enabled.load(Ordering::Relaxed) && failsafe {
```

crates/relay/src/auctioneer/context.rs
```diff
@@ -133,6 +183,10 @@ impl<B: BidAdjustor> Context<B> {
                     {
                         failsafe_triggered.store(true, Ordering::Relaxed);
                         error!(%builder, %err, %block_hash, "failed to demote builder in database! Pausing all optmistic submissions");
+                        alert_discord(&format!(
+                            "{} {} {} failed to demote builder in database! Pausing all optmistic submissions",
+                            builder, err, block_hash
+                        ));
                     }
                 });
             } else {
```
ltitanb, line 196: this else should be updated as well if it was an adjusted block

crates/relay/src/auctioneer/mod.rs
```diff
@@ -338,6 +346,29 @@ impl State {
                 ctx.handle_simulation_result((id, None));
             }
 
+            // Dry run adjusting & validating bids
```
gd-0: lets add a bit more detail to this comment explaining why. 

- validate bids throughout the slot
- aiming to catch any adjustment bugs early
- means we hopefully stop doing adjustments in get_header before the actual proposer call is made

crates/relay/src/auctioneer/context.rs
```diff
@@ -196,8 +250,8 @@ impl<B: BidAdjustor> Context<B> {
             return;
         };
 
-        let original_payload_and_blobs = original_payload.payload_and_blobs.clone();
-        let builder_pubkey = original_payload.bid_data.builder_pubkey;
+        let original_payload_and_blobs = original_payload.payload_and_blobs();
+        let builder_pubkey = original_payload.bid_data().builder_pubkey;
```
ltitanb: here we are cloning the full `BidData` just to get the builder pubkey, let's return a ref instead

crates/relay/src/auctioneer/context.rs
```diff
@@ -48,11 +51,16 @@ pub struct Context<B: BidAdjustor> {
...
     pub bid_adjustor: B,
+    pub adjustments_enabled: Arc<AtomicBool>,
+    pub adjustments_failsafe_trigger: Arc<AtomicBool>,
 }
 
 const EXPECTED_PAYLOADS_PER_SLOT: usize = 5000;
 const EXPECTED_BUILDERS_PER_SLOT: usize = 200;
 
+const DB_CHECK_INTERVAL_SEC: u64 = 5;
+const ADJUSTMENTS_DRY_RUN_INTERVAL_SEC: u64 = 2;
```
gd-0: auction window can be quite tight sometimes. would lower this to something like every 500ms. load is not really a concern here

crates/relay/src/auctioneer/mod.rs
```diff
@@ -338,6 +346,29 @@ impl State {
...
+                    );
+                    return;
+                };
+
+                if let Some(adjusted_bid) =
+                    ctx.bid_adjustor.try_apply_adjustments(original_bid, true)
+                {
+                    if let Some(sim_request) = adjusted_bid.into_sim_request() {
+                        ctx.sim_manager.handle_sim_request(sim_request);
+                    }
```
gd-0: i think the sim manager has a queue system underneath. maybe its worth adding a fast track lane for these kind of checks to make sure they don't get dropped?

crates/relay/src/auctioneer/get_header.rs
```diff
@@ -33,11 +35,18 @@ impl<B: BidAdjustor> Context<B> {
...
+            if let Some(adjusted_bid) = self.bid_adjustor.try_apply_adjustments(original_bid, false)
+            {
+                let block_hash = adjusted_bid.block_hash();
+                self.payloads.insert(*block_hash, adjusted_bid.clone());
+
+                if let Some(sim_request) = adjusted_bid.into_sim_request() {
```
ltitanb: since whenever we call the `try_apply_adjustments` we then do a `into_sim_request` i think a better api is for the `try_apply_adjustments` to return both the adjusted_bid and the sim_request

crates/relay/src/auctioneer/mod.rs
```diff
@@ -338,6 +346,29 @@ impl State {
                 ctx.handle_simulation_result((id, None));
             }
 
+            // Dry run adjusting & validating bids
+            (_, Event::DryRunAdjustments) => {
```
ltitanb: why not send a `GetHeader` event possibly with a flag to skip some checks? that way you ensure the state and flow are similar to the actual prod flow

crates/relay/src/auctioneer/types.rs
```diff
@@ -237,6 +326,23 @@ pub struct SlotData {
...
+pub struct SimSlotData {
+    pub gas_limit: u64,
+    pub validator_preferences: ValidatorPreferences,
+    pub inclusion_list: Option<InclusionListWithMetadata>,
+}
+
+impl From<&SlotData> for SimSlotData {
+    fn from(data: &SlotData) -> Self {
+        Self {
+            gas_limit: data.registration_data.entry.registration.message.gas_limit.clone(),
```
ltitanb: we should have a clippy check for this but if the data is `Copy` we can avoid the `.clone()`

crates/relay/src/auctioneer/get_header.rs
ltitanb: that way we can avoid the `SimSlotData` field entirely

crates/types/src/bid_submission.rs
```diff
@@ -325,6 +325,15 @@ impl SignedBidSubmission {
         }
     }
 
+    pub fn message_mut(&mut self) -> &mut BidTrace {
```
ltitanb, line 328: this is never used?

crates/types/src/bid_submission.rs
```diff
@@ -336,15 +345,26 @@ impl SignedBidSubmission {
         }
     }
 
-    pub fn payload_and_blobs_ref(&self) -> PayloadAndBlobsRef<'_> {
+    pub fn execution_payload_make_mut(&mut self) -> &mut ExecutionPayload {
```
ltitanb, line 348: same here, let's remove if not used

ltitanb (changes requested):

crates/relay/src/auctioneer/mod.rs
ltitanb: agreed i think we should ideally keep a different queues for adjusted vs not bids

crates/relay/src/auctioneer/mod.rs
ninaiiad: actually exactly because having to match the valid state would block many dry runs and the adjustments themselves don't really depend on that
also because we don't know the parent hash for the block we want to adjust

crates/relay/src/auctioneer/context.rs
ninaiiad: true, though I think it might be better to add an extra update so it can't get overwritten and then only check this flag https://github.com/gattaca-com/helix/blob/0a2ceb0cbb7b25e544ede3a7cceca3895ef23da9/crates/relay/src/auctioneer/context.rs#L163-L165

crates/relay/src/auctioneer/types.rs
ninaiiad: the problem is that for one of the match arms this struct is assembled in-place
and this ref trick (https://github.com/gattaca-com/helix/blob/develop/crates/types/src/blobs.rs#L417-L421) I think wouldn't really help much because it's a struct with two references (to Arcs) vs struct with two Arcs

crates/relay/src/auctioneer/simulator/manager.rs
```diff
@@ -58,6 +58,7 @@ pub struct SimulationResultInner {
 pub struct SimulatorManager {
     simulators: Vec<SimulatorClient>,
     requests: PendingRquests,
+    priority_requests: PendingRquests,
```
ltitanb, line 61: we need to clear this in the `on_new_slot` below

crates/relay/src/auctioneer/context.rs
```diff
@@ -214,6 +272,60 @@ impl<B: BidAdjustor> Context<B> {
...
+                    Ok(value) => {
+                        let previous = flag.swap(value, Ordering::Relaxed);
+                        if previous != value {
+                            tracing::info!(
+                                "adjustments enabled flag changed from {} to {}",
+                                previous,
+                                value
+                            );
+                        }
+                        flag.store(value, Ordering::Relaxed)
```
ltitanb: this is unneeded since you did a `.swap` above

crates/relay/src/auctioneer/context.rs
```diff
@@ -214,6 +272,60 @@ impl<B: BidAdjustor> Context<B> {
...
+        db: Arc<PostgresDatabaseService>,
+        flag: Arc<AtomicBool>,
+        failsafe_triggered: Arc<AtomicBool>,
+    ) {
+        spawn_tracked!(async move {
+            let mut interval = tokio::time::interval(Duration::from_secs(DB_CHECK_INTERVAL_SEC));
+            loop {
+                interval.tick().await;
+
+                if failsafe_triggered.load(Ordering::Relaxed) {
```
ltitanb, line 286: we never store `false` here, so if the failsafe is ever triggered we would need a relay restart to reset it? got me thinking, is there much value in having the enabled vs failsafe flags separate?

crates/relay/src/auctioneer/mod.rs
ltitanb: i don't think that's right, why would you try adjust something when you don't have blocks yet or are already broadcasting a given slot? i think makes sense to try adjust only in `Sorting` state. You could add a flag to skip the parent hash check

crates/relay/src/auctioneer/context.rs
ninaiiad: https://github.com/gattaca-com/helix/pull/350#discussion_r2657471868

ltitanb: makes sense , looks like for optimistic submissions as well we never revert the failsafe

crates/relay/src/auctioneer/mod.rs
```diff
@@ -338,6 +345,28 @@ impl State {
...
+                        "adjustments dry run - failed to get payload from bid sorter, this should never happen!"
+                    );
+                    return;
+                };
+
+                if let Some((_, sim_request)) =
+                    ctx.bid_adjustor.try_apply_adjustments(original_bid, slot_data, true)
+                {
+                    ctx.sim_manager.handle_sim_request(sim_request, true);
+                }
```
ltitanb, line 367: is it useful to add some logs here maybe , to print some info on what we adjusted?

crates/relay/src/auctioneer/context.rs
```diff
@@ -48,11 +51,16 @@ pub struct Context<B: BidAdjustor> {
...
     pub bid_adjustor: B,
+    pub adjustments_enabled: Arc<AtomicBool>,
+    pub adjustments_failsafe_trigger: Arc<AtomicBool>,
 }
 
 const EXPECTED_PAYLOADS_PER_SLOT: usize = 5000;
 const EXPECTED_BUILDERS_PER_SLOT: usize = 200;
 
+const DB_CHECK_INTERVAL_SEC: u64 = 1;
+const ADJUSTMENTS_DRY_RUN_INTERVAL_MS: u64 = 500;
```
ltitanb: nit you can declare a `Duration` const so you dont have to specify the unit

ltitanb (approved):

crates/relay/src/auctioneer/mod.rs
ninaiiad: there will be logs in the implementation

crates/relay/src/auctioneer/context.rs
```diff
@@ -117,7 +147,31 @@ impl<B: BidAdjustor> Context<B> {
...
+                    SimulatorMetrics::disable_adjustments();
+                    self.adjustments_enabled.store(false, Ordering::Relaxed);
+
+                    let db = self.db.clone();
+                    let failsafe_trigger = self.adjustments_failsafe_trigger.clone();
+                    let adjustments_enabled = self.adjustments_enabled.clone();
+                    spawn_tracked!(async move {
+                        if let Err(err) = db.disable_adjustments().await {
+                            failsafe_trigger.store(true, Ordering::Relaxed);
+                            adjustments_enabled.store(false, Ordering::Relaxed);
```
gd-0, line 165: lets set adjustments_enabled before spawning just got get it propagated faster.

ninaiiad: it is set before spawning, this is to ensure that the background task does not update it to true by mistake (it only can't do that after the failsafe_trigger is set as well)

PR #351, opened 2026-01-06
fix for missing bid submissions

0w3n-d:
Currently we save to the db when we get the sim result back, however sims can be dropped or timeout and the task is then dropped so no result and nothing saved in the db.

### DIFF
--- a/crates/relay/src/auctioneer/submit_block.rs
+++ b/crates/relay/src/auctioneer/submit_block.rs
@@ -39,7 +39,8 @@ impl<B: BidAdjustor> Context<B> {
                     Some(res_tx)
                 };
 
-                self.simulate_and_store(validated, res_tx, slot_data);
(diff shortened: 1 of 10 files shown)


crates/relay/src/auctioneer/submit_block.rs
```diff
@@ -232,7 +230,34 @@ impl<B: BidAdjustor> Context<B> {
...
+        let sub_clone = req.submission.clone();
+        let trace_clone = req.trace.clone();
+        let opt_version = req.optimistic_version();
+
+        let db = self.db.clone();
+        spawn_tracked!(async move {
+            if let Err(err) =
+                db.store_block_submission(sub_clone, trace_clone, opt_version, is_adjusted).await
+            {
+                error!(%err, "failed to store block submission")
```
ninaiiad, line 256: let's also log the block_hash here?

ninaiiad (changes requested):

PR #353, opened 2026-01-06
fix Unknown registrations return 200 instead of 400 error response

0ex-d:
For more information, please refer to this [issue](https://github.com/gattaca-com/helix/issues/191).


0ex-d:
Note: This covers at-least-one unknown registration.

PR #355, opened 2026-01-08
Od/tile prep

0w3n-d:
Picking up from where we left off with the auctioneer work.
This is the first of a number planned PRs to isolate all IO, async, tokio, and so on away from the rest of the code.
The next step then will be to move any remaining non IO dependent code out of these async services into regular threads.
Leaving the async services as very thin wrappers around IO crates like axum, postgres, and so on.
The non async code can then be restructured into tiles.
The final step will be to replace the small set of async tasks remaining with custom networking code.

---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/355" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/relay/src/database/postgres/postgres_db_service.rs
+++ b/crates/relay/src/database/postgres/postgres_db_service.rs
@@ -1079,7 +1079,7 @@ impl PostgresDatabaseService {
     #[instrument(skip_all)]
     async fn fetch_validator_registrations_since(
         &self,
-        last_fetch_time: SystemTime,
+        let mut record = DbMetricRecord::new("fetch_validator_registrations_since");
     ) -> Result<Vec<SignedValidatorRegistrationEntry>, DatabaseError> {
         let mut record = DbMetricRecord::new("get_validator_registrations");
 


crates/common/src/local_cache.rs
```diff
@@ -238,6 +264,58 @@ impl LocalCache {
...
     }
+
+    pub fn is_registration_update_required(
+        &self,
+        registration: &SignedValidatorRegistration,
+    ) -> bool {
+        if let Some(existing_entry) =
+            self.validator_registration_cache.get(&registration.message.pubkey) &&
+            existing_entry.registration_info.registration.message.timestamp >=
+                registration.message.timestamp.saturating_sub(60 * 60) &&
```
gd-0: can we make this a const? Tink it's always easier to read.
Doesn;'t need to be global just can define the const as the first line  of the fn

crates/relay/src/database/handle.rs
```diff
@@ -0,0 +1,214 @@
...
+use crate::database::{
+    BuilderInfoDocument, SavePayloadParams,
+    postgres::postgres_db_service::{DbRequest, PendingBlockSubmissionValue},
+};
+
+#[derive(Clone)]
+pub struct DbHandle {
+    sender: crossbeam_channel::Sender<DbRequest>,
+    batch_sender: crossbeam_channel::Sender<PendingBlockSubmissionValue>,
+}
```
gd-0, line 30: lets add a comment to this just explaining that this wrapper around channels is just temporary as we work on incremental changes towartds the flux design

crates/relay/src/database/postgres/postgres_db_service.rs
```diff
@@ -228,9 +305,71 @@ impl PostgresDatabaseService {
...
+
+    pub async fn start_db_request_processor(
+        &mut self,
+        receiver: crossbeam_channel::Receiver<DbRequest>,
+    ) {
+        let svc_clone = self.clone();
+
+        tokio::spawn(async move {
+            loop {
+                match receiver.recv() {
```
gd-0: how does this work in tokio btw? IIRC it thread::yields so should be ok. Just not sure if it's yielding to the tokio runtime properly.

gd-0: if this is an issue could loop through try_recv until the queue is empty and then tokio::sleep for a short time

gd-0 (approved):
looks good. just some minor comments

crates/relay/src/housekeeper/housekeeper.rs
```diff
@@ -47,12 +45,9 @@ struct HousekeeperSlots {
     head: Arc<AtomicU64>,
     proposer_duties: Arc<AtomicU64>,
     refreshed_validators: Arc<AtomicU64>,
-    trusted_proposers: Arc<AtomicU64>,
     // updating can take >> slot time, so we avoid concurrent updates by locking the mutex
-    updating_builder_infos: Arc<Mutex<()>>,
     updating_proposer_duties: Arc<Mutex<()>>,
```
ninaiiad, line 49: is this still needed?

crates/relay/src/database/handle.rs
```diff
@@ -0,0 +1,215 @@
...
+};
+use helix_types::{BlsPublicKeyBytes, MergedBlock, SignedBidSubmission};
+use tracing::error;
+
+use crate::database::{
+    BuilderInfoDocument, SavePayloadParams,
+    postgres::postgres_db_service::{DbRequest, PendingBlockSubmissionValue},
+};
+
+// This is temporary until we have the spine connected up
```
ninaiiad, line 25: we do have the spine now - or do you think it would be simpler to use crossbeam for now still and change later?

crates/relay/src/database/postgres/postgres_db_service.rs
```diff
@@ -236,9 +327,95 @@ impl PostgresDatabaseService {
...
+    }
+
+    pub async fn start_db_request_processor(
+        &mut self,
+        receiver: crossbeam_channel::Receiver<DbRequest>,
+    ) {
+        let svc_clone = self.clone();
+
+        tokio::spawn(async move {
+            loop {
```
ninaiiad, line 401: should we put this into a tile?

crates/relay/src/database/postgres/postgres_db_service.rs
```diff
@@ -236,9 +327,95 @@ impl PostgresDatabaseService {
...
         self.start_registration_processor().await;
-        self.start_block_submission_processor().await;
+        self.start_block_submission_processor(block_submission_receiver).await;
+        self.start_db_request_processor(db_request_receiver).await;
+        self.start_check_flag_task();
+    }
+
+    fn start_check_flag_task(&mut self) {
+        let svc_clone = self.clone();
+        tokio::spawn(async move {
```
ninaiiad, line 364: do you think if we decide to move away from tokio completely we could just put this inside the loop_body with some wrapper that would check elapsed time? and if this isn't the only db reload check, we could maybe merge it into one request (& table)?

crates/common/src/local_cache.rs
```diff
@@ -228,6 +245,57 @@ impl LocalCache {
...
+                registration
+                    .message
+                    .timestamp
+                    .saturating_sub(VALIDATOR_REGISTRATION_UPDATE_INTERVAL) &&
+            existing_entry.registration_info.registration.message.fee_recipient ==
+                registration.message.fee_recipient &&
+            existing_entry.registration_info.registration.message.gas_limit ==
+                registration.message.gas_limit
+        {
+            // do registration once per hour, unless fee recipient / gas limit has changed
```
ninaiiad: is this a todo?

crates/relay/src/housekeeper/housekeeper.rs
0w3n-d: maybe not, the next thing I'm moving on to is to refactor the housekeeper and move it to a tile so if I can remove it I'll do it that PR

crates/relay/src/database/handle.rs
0w3n-d: Yeah I wrote this PR weeks ago before the tiles were added. It's a pretty big PR now.
I think I'll just get it merged and work on moving stuff to tiles in a follow up PR.

crates/relay/src/database/postgres/postgres_db_service.rs
0w3n-d: yeah will do in follow up PR

crates/relay/src/database/postgres/postgres_db_service.rs
0w3n-d: Yeah probably. The current approach is certainly far from optimal.

crates/common/src/local_cache.rs
0w3n-d: I'll remove this. I can't remember what that comment was about. I think I had written a different way but then changed it and forgot to remove the comment.

PR #362, opened 2026-01-12
dry run get_header

ninaiiad:


crates/relay/src/auctioneer/get_header.rs
```diff
@@ -40,15 +40,18 @@ impl<B: BidAdjustor> Context<B> {
...
         if self.adjustments_enabled.load(Ordering::Relaxed) &&
-            let Some((adjusted_bid, sim_request)) =
-                self.bid_adjustor.try_apply_adjustments(original_bid, slot_data, false)
+            let Some((adjusted_bid, sim_request, is_adjustable_slot)) =
+                self.bid_adjustor.try_apply_adjustments(&original_bid, slot_data, false)
         {
             self.store_data_and_sim(sim_request, adjusted_bid.clone(), true);
 
-            return Ok(adjusted_bid);
+            if is_adjustable_slot {
```
vladimir-ea, line 50: should the 'store' line above be moved inside the check for the adjustable slot? is there a reason to store otherwise?

ninaiiad: this repeats the dry run mode, so it's okay. the important part is simulation

PR #382, opened 2026-02-06
tile spinloops

ninaiiad:


crates/relay/src/main.rs
```diff
@@ -85,88 +88,133 @@ async fn run(instance_id: String, config: RelayConfig, keypair: BlsKeypair) -> e
...
-    let mut sigint = tokio::signal::unix::signal(SignalKind::interrupt())?;
-    let mut sigterm = tokio::signal::unix::signal(SignalKind::terminate())?;
+    let termination_grace_period = config.router_config.shutdown_delay_ms;
 
-    tokio::select! {
-        _ = sigint.recv() => {}
-        _ = sigterm.recv() => {}
-    }
+    let spine = HelixSpine::new(None);
+    // TODO: on panic?
```
ltitanb: lets move the discord alert here

ninaiiad: it will be inherited from the panic hook set earlier

PR #386, opened 2026-02-13
tcp tile for bid submissions

ninaiiad:
---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/386" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,10 +42,10 @@ ethers = "2.0.14"
 eyre = "0.6.12"
 flate2 = "1.0"
 
-flux = { git = "https://github.com/gattaca-com/flux", branch = "ng/reconnect"}
-flux-network = { package = "flux-network", git = "https://github.com/gattaca-com/flux", branch = "ng/reconnect"}
-flux-type-hash = { package = "type-hash", git = "https://github.com/gattaca-com/flux",branch = "ng/reconnect"}
-flux-type-hash-derive = { package = "type-hash-derive", git = "https://github.com/gattaca-com/flux",branch = "ng/reconnect"}
+flux = { git = "https://github.com/gattaca-com/flux", rev = "15dfc44b360c3ec55abdf8e506f2dca60ac72e13"}
+flux-network = { package = "flux-network", git = "https://github.com/gattaca-com/flux", rev = "15dfc44b360c3ec55abdf8e506f2dca60ac72e13"}
+flux-type-hash = { package = "type-hash", git = "https://github.com/gattaca-com/flux", rev = "15dfc44b360c3ec55abdf8e506f2dca60ac72e13"}
+flux-type-hash-derive = { package = "type-hash-derive", git = "https://github.com/gattaca-com/flux", rev = "15dfc44b360c3ec55abdf8e506f2dca60ac72e13"}
 
 futures = "0.3"
 futures-util = { version = "0.3", features = ["compat"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4533,7 +4533,7 @@ dependencies = [
 [[package]]
 name = "flux"
 version = "0.0.37"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "bitcode",
  "core_affinity",
@@ -4556,7 +4556,7 @@ dependencies = [
 [[package]]
 name = "flux-communication"
 version = "0.0.26"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "directories",
  "flux-timing",
@@ -4570,7 +4570,7 @@ dependencies = [
 [[package]]
 name = "flux-network"
 version = "0.0.8"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "flux-communication",
  "flux-timing",
@@ -4583,7 +4583,7 @@ dependencies = [
 [[package]]
 name = "flux-timing"
 version = "0.0.15"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "bitcode",
  "chrono",
@@ -4601,7 +4601,7 @@ dependencies = [
 [[package]]
 name = "flux-utils"
 version = "0.0.22"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "core_affinity",
  "directories",
@@ -12931,7 +12931,7 @@ checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
 [[package]]
 name = "spine-derive"
 version = "0.0.9"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -14261,12 +14261,12 @@ dependencies = [
 [[package]]
 name = "type-hash"
 version = "0.0.1"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 
 [[package]]
 name = "type-hash-derive"
 version = "0.0.4"
-source = "git+https://github.com/gattaca-com/flux?branch=ng%2Freconnect#b4c209f6ce592dcc3397a9d68921be01d56f5741"
+source = "git+https://github.com/gattaca-com/flux?rev=15dfc44b360c3ec55abdf8e506f2dca60ac72e13#15dfc44b360c3ec55abdf8e506f2dca60ac72e13"
 dependencies = [
  "proc-macro-crate",
  "proc-macro2",


crates/relay/src/tcp_bid_recv/mod.rs
ninaiiad, line 1: tcp tile

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,189 @@
...
+        tracing::trace!("{:?}", header);
+
+        let bid_submission_data = Bytes::copy_from_slice(&data[BID_SUB_HEADER_SIZE..]);
+        Ok(BidSubmission { header, data: bid_submission_data })
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy, Default)]
+pub struct BidSubmissionHeader {
```
ninaiiad: the bid submission header (builder would send this + the ssz encoded bid)

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,189 @@
...
+            BidSubmissionError::InvalidMergeType(_) => Status::InvalidRequest,
+            BidSubmissionError::SubmissionTooShort => Status::InvalidRequest,
+            // _ => Status::InvalidRequest,
+        }
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Clone, Encode)]
+pub struct BidSubmissionResponse {
```
ninaiiad: the response from relay to the builder. the sequence number is used as a unique identifier for a bid submission

crates/relay/src/tcp_bid_recv/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+                                .is_some_and(|p| p.value().contains(&msg.builder_pubkey))
+                            {
+                                self.registered.insert(token);
+                            } else {
+                                tracing::error!(
+                                    "unknown api key and pubkey pair: {} {}, disconnecting peer",
+                                    api_key,
+                                    msg.builder_pubkey
+                                );
+                                self.to_disconnect.push(token);
```
ninaiiad, line 166: using an extra vector because we can't borrow the TcpConnector listener instance as mutable in this closure

crates/relay/src/main.rs
```diff
@@ -174,6 +175,22 @@ async fn run(instance_id: String, config: RelayConfig, keypair: BlsKeypair) -> e
             }
 
             let auctioneer_core = config.cores.auctioneer;
+            let sock_addr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, config.tcp_port));
```
0w3n-d: This needs to be `Ipv4Addr::UNSPECIFIED` so it can receive external connections.

crates/common/src/config.rs
```diff
@@ -89,9 +93,12 @@ impl RelayConfig {
                 tokio: vec![],
                 sub_workers: vec![],
                 reg_workers: vec![],
+                tcp_bid_submissions_tile: 2,
             },
             gossip_payload_on_header: false,
             api_port: 4040,
+            tcp_port: 4041,
+            tcp_max_connections: Default::default(),
```
ltitanb: this makes `tcp_max_connections` 0 by default , vs 512 from the serde default

Cargo.toml
```diff
@@ -41,6 +42,7 @@ ethers = "2.0.14"
 eyre = "0.6.12"
 flate2 = "1.0"
 flux = { git = "https://github.com/gattaca-com/flux" }
+flux-network = { package = "flux-network", git = "https://github.com/gattaca-com/flux" }
```
ltitanb: would make sure to pin the tag here

crates/common/src/local_cache.rs
```diff
@@ -89,7 +88,7 @@ pub struct LocalCache {
     pub inclusion_list: Arc<RwLock<Option<InclusionListWithKey>>>,
     builder_info_cache: Arc<DashMap<BlsPublicKeyBytes, BuilderInfo>>,
     /// Api key -> builder pubkey
-    api_key_cache: Arc<DashMap<HeaderValue, Vec<BlsPublicKeyBytes>>>,
+    pub api_key_cache: Arc<DashMap<String, Vec<BlsPublicKeyBytes>>>,
```
ltitanb, line 91: why was this necessary? the previous version avoided one hop of parsing / allocation

crates/relay/src/main.rs
```diff
@@ -121,7 +122,7 @@ async fn run(instance_id: String, config: RelayConfig, keypair: BlsKeypair) -> e
     let termination_grace_period = config.router_config.shutdown_delay_ms;
```
ltitanb: nit and unrelated to this PR but would rather create the `Duration` here as we are losing the _ms

crates/relay/src/auctioneer/types.rs
```diff
@@ -27,12 +27,17 @@ use tracing::debug;
...
+        decoder::{Compression, Encoding},
+        simulator::manager::SimulationResult,
+    },
     gossip::BroadcastPayloadParams,
     housekeeper::PayloadAttributesUpdate,
+    tcp_bid_recv::types::BidSubmissionHeader,
 };
 
-pub type SubmissionResult = Result<(), BuilderApiError>;
+pub type SubmissionResult = (u64, Result<(), BuilderApiError>);
```
ltitanb: unclear what this new `u64`? would either add a comment or new type if you are using it in several places

crates/relay/src/api/builder/submit_block.rs
```diff
@@ -26,18 +29,33 @@ impl<A: Api> BuilderApi<A> {
...
+        tracing::Span::current().record("id", tracing::field::display(request_id));
+
         trace!("start handler");
 
         let mut trace = SubmissionTrace::init_from_timings(timings);
         trace.metadata = api.api_provider.get_metadata(&headers);
 
-        let Ok(rx) = api.auctioneer_handle.block_submission(headers, body, trace) else {
+        let (header, sequence, encoding, compression, api_key) =
+            headers_map_to_bid_submission_header(headers);
```
ltitanb: how about grouping all of these in a single struct that you can create from either the headersMap or the tcp sub headers?

crates/relay/src/api/builder/submit_block.rs
```diff
@@ -26,18 +29,33 @@ impl<A: Api> BuilderApi<A> {
         headers: HeaderMap,
         body: bytes::Bytes,
     ) -> Result<(), BuilderApiError> {
+        let request_id = extract_request_id(&headers);
```
ltitanb, line 32: `extract_request_id` returns a `Uuid` but in a few other places we now have `request_id` as u64. Would be good to clarify the discrepancy or use the same type if possible

crates/relay/src/auctioneer/decoder.rs
```diff
@@ -299,6 +346,70 @@ fn gzip_size_hint(buf: &[u8]) -> Option<usize> {
...
+    }
+
+    if let Some(submission_type) = SubmissionType::from_headers(&headers) {
+        match submission_type {
+            SubmissionType::Dehydrated => {
+                flags.set(BidSubmissionFlags::IS_DEHYDRATED, true);
+            }
+            SubmissionType::Merge => {
+                // TODO @nina
+                // flags.set(BidSubmissionFlags::WITH_MERGEABLE_DATA, true);
```
ltitanb: todo before merging

crates/relay/src/auctioneer/mod.rs
```diff
@@ -138,12 +140,18 @@ impl Default for State {
 // TODO: tokio metrics
 
 impl State {
-    fn step<B: BidAdjustor>(&mut self, event: Event, ctx: &mut Context<B>, tel: &mut Telemetry) {
+    fn step<B: BidAdjustor>(
+        &mut self,
+        event: Event,
+        ctx: &mut Context<B>,
+        tel: &mut Telemetry,
+        _adapter: &mut flux::spine::SpineAdapter<HelixSpine>,
```
ltitanb: why pass the adapter if it's unused?

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,189 @@
...
+use crate::{
+    api::builder::error::BuilderApiError,
+    auctioneer::{Compression, MergeType},
+};
+
+#[repr(C)]
+#[derive(Debug, Clone, Decode)]
+pub struct RegistrationMsg {
+    pub api_key: [u8; 16], // this is a Uuid
+    pub builder_pubkey: BlsPublicKeyBytes,
```
ltitanb: we should double check that every submission from this token has the correct public key since we always skip sigverify

crates/relay/src/tcp_bid_recv/types.rs
ltitanb: worth adding this in docs that makes it clearer what is the builder sending and how's it different than the http submission (doesn't need to be super detailed)

ltitanb: you should double check alignment here, in general better adding some padding at the end

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,189 @@
...
+        if value.len() < BID_SUB_HEADER_SIZE {
+            return Err(BidSubmissionError::SubmissionTooShort)
+        }
+
+        let sequence_number = u64::from_be_bytes(value[..8].try_into().unwrap());
+        let merge_type = MergeType::try_from(value[8])
+            .map_err(|_| BidSubmissionError::InvalidMergeType(value[8]))?;
+        let flags = BidSubmissionFlags::from_bits_retain(value[9]);
+
+        Ok(Self { sequence_number, merge_type, flags })
```
ltitanb: let's add a test that encodes and decodes. If properly structured this can also be imported directly from the builder side instead of copy pasting. Also would add a test for
```rust
assert_eq!(size_of::<BidSubmissionHeader>(), BID_SUB_HEADER_SIZE)
```

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,189 @@
...
+        let bid_submission_data = Bytes::copy_from_slice(&data[BID_SUB_HEADER_SIZE..]);
+        Ok(BidSubmission { header, data: bid_submission_data })
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy, Default)]
+pub struct BidSubmissionHeader {
+    pub sequence_number: u64,
+    pub merge_type: MergeType,
```
ltitanb: why this needs it's own field, can't be in `BidSubmissionFlags` ?

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+    pub fn with_adjustments(&self) -> bool {
+        self.contains(Self::WITH_ADJUSTMENTS)
+    }
+    pub fn is_zstd_compressed(&self) -> bool {
+        self.contains(Self::ZSTD_COMPRESSED)
+    }
+}
+
+impl Decode for BidSubmissionFlags {
```
ltitanb: this flags should not be ssz encoded, if we are handrolling the conversion above

crates/relay/src/tcp_bid_recv/types.rs
ltitanb: better to always put variable length fields after fixed length fields

crates/relay/src/tcp_bid_recv/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+                            .ssz_append(buffer);
+                    });
+                    self.to_remove.push(*token);
+                }
+                Err(err) if err == TryRecvError::Closed => {
+                    self.to_remove.push(*token);
+                }
+                _ => {}
+            }
+        }
```
ltitanb: this is not ideal, why not use a single channel were the results are pushed? that way you can just check if anything has been pushed instead of looping over a hashmap. You can pass a ref to the return message so you know which token to send to

ltitanb (changes requested):

crates/common/src/local_cache.rs
ninaiiad: because I check this map from the tcp tile too

crates/relay/src/tcp_bid_recv/types.rs
ninaiiad: this will have implicit padding, but I don't think that matters if we decode / encode manually? doesn't need a repr(C) either actually

crates/relay/src/tcp_bid_recv/types.rs
ninaiiad: as an enum it stores values that are mutually exclusive - this is cleaner than having to manually check if two incompatible bits are set at once

crates/relay/src/api/builder/submit_block.rs
ninaiiad: will change the seq num id to u128 and parse it as uuid for tagging in traces, then it's the same

crates/relay/src/tcp_bid_recv/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+                );
+            },
+            |token, msg, _| {
+                if let Some(expected_pubkey) = self.registered.get(&token) {
+                    match BidSubmission::try_from(msg) {
+                        Ok(bid) => {
+                            let request_id = bid.header.sequence_number;
+                            match Self::accept_bid_submission(self.tx.clone(), &self.auctioneer_handle, bid, *expected_pubkey) {
+                                Ok(_) => {
+                                    self.in_flight.insert(request_id, token);
```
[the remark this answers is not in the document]
ninaiiad: could be okay if we expect that this u128 is a representation of some (random?) uuid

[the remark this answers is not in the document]
ltitanb: why not send back the token with the response to the tile? that way you dont even need the map

crates/relay/src/api/builder/top_bid.rs
```diff
@@ -22,7 +22,10 @@ impl<A: Api> BuilderApi<A> {
         headers: HeaderMap,
         ws: WebSocketUpgrade,
     ) -> Result<impl IntoResponse, BuilderApiError> {
-        let Some(api_key) = headers.get(HEADER_API_KEY).or_else(|| headers.get(HEADER_API_TOKEN))
+        let Some(api_key) = headers
+            .get(HEADER_API_KEY)
+            .or(headers.get(HEADER_API_TOKEN))
```
ltitanb: would still use or_else since it's lazy

crates/relay/src/auctioneer/types.rs
```diff
@@ -27,15 +28,53 @@ use tracing::debug;
...
+        }
+    }
+
+    pub fn is_closed(&self) -> bool {
+        match self {
+            Self::OneShot(tx) => tx.is_closed(),
+            Self::Shared(_) => false,
+        }
+    }
+}
```
ltitanb, line 177: a side effect of this is that we will add some load to the simulators since we would never drop pending sims. Not a problem but worth adding a short comment on this

crates/relay/src/tcp_bid_recv/types.rs
```diff
@@ -0,0 +1,218 @@
...
+            BidSubmissionError::InvalidMergeType(_) => Status::InvalidRequest,
+            BidSubmissionError::SubmissionTooShort => Status::InvalidRequest,
+            // _ => Status::InvalidRequest,
+        }
+    }
+}
+
+/// SSZ-encoded server response sent back in a single frame.
+#[repr(C)]
+#[derive(Debug, Clone, Encode)]
```
ltitanb: fine either way but not sure it adds much to ssz encode it?

ninaiiad: not really, but if it were ever extended, especially with ethereum-specific types like hashes and keys, this is cleaner

crates/relay/src/auctioneer/types.rs
ninaiiad: those queues are cleaned up on every new slot, so should be ok

crates/relay/src/auctioneer/decoder.rs
```diff
@@ -299,6 +283,60 @@ fn gzip_size_hint(buf: &[u8]) -> Option<usize> {
...
+
+    let merge_type = merge_type_from_headers(&headers, submission_type);
+
+    let api_key = headers
+        .get(HEADER_API_KEY)
+        .or(headers.get(HEADER_API_TOKEN))
+        .and_then(|key| key.to_str().map(|key| key.to_owned()).ok());
+
+    let header =
+        BidSubmissionHeader { sequence_number: seq_num.unwrap_or_default(), merge_type, flags };
```
louisponet: Should we have an error code path here rather than the unwrap_or_default ? Idk if it's a footgun

crates/relay/src/auctioneer/types.rs
```diff
@@ -27,15 +29,49 @@ use tracing::debug;
...
+    auctioneer::{BlockMergeResult, decoder::Encoding, simulator::manager::SimulationResult},
     gossip::BroadcastPayloadParams,
     housekeeper::PayloadAttributesUpdate,
+    tcp_bid_recv::BidSubmissionHeader,
 };
 
-pub type SubmissionResult = Result<(), BuilderApiError>;
+#[derive(Clone, Copy, Debug)]
+pub struct SubmissionRef {
+    pub seq_num: Option<SeqNum>,
```
louisponet: Are both Some in case the submission came from a tcp? In that case I think it's better to have Option<SubmissionTcp> (I like that name since it's a bit more explicit) or so in the SubmissionResult.

crates/relay/src/auctioneer/worker.rs
```diff
@@ -231,70 +250,61 @@ impl SubWorker {
     #[allow(clippy::type_complexity)]
     fn handle_block_submission(
```
louisponet, line 239: Feels like there could be a substruct in the BlockSubmission one with these arguments?

crates/relay/src/tcp_bid_recv/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+        let listener = TcpConnector::default()
+            .with_telemetry(TcpTelemetry::Enabled { app_name: HelixSpine::app_name() })
+            .with_socket_buf_size(64 * 1024 * 1024); // 64MB
+
+        let (tx, rx) = crossbeam_channel::bounded(10_000);
+
+        Self {
+            listener_addr,
+            listener,
+            listener_token: None,
```
louisponet: I think we can make this a requirement? I.e. we do listen_at here and do .expect on the token as I don't think we should even start up if the tcp listening is broken

PR #409, opened 2026-03-03
bid submissions refactor

ninaiiad:
Key changes:
- Encoded bid submissions are stored in a shared vec + their indices are sent to decoder workers via a spine queue
- Bid submission responses are too stored in a shared vec
    - the Tcp tile consumes new results from the queue directly
    - the axum http handlers await on futures in another vec which is populated by a fan-out tile

flux pr dependency: https://github.com/gattaca-com/flux/pull/37

---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/409" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/common/src/traces/builder_api_trace.rs
```diff
@@ -1,19 +1,19 @@
...
 use serde::{Deserialize, Serialize};
 
 use crate::{RequestTimings, metrics::SUB_TRACE_LATENCY, utils::utcnow_ns};
 
-// all timestamps are in nanoseconds
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct SubmissionTrace {
     // first packet
     pub receive_ns: u64,
```
vladimir-ea: should we change timestamps to use `Nanos` type?

crates/relay/src/api/builder/api.rs
```diff
@@ -18,6 +25,10 @@ pub struct BuilderApi<A: Api> {
     pub top_bid_tx: tokio::sync::broadcast::Sender<TopBidUpdate>,
     pub auctioneer_handle: AuctioneerHandle,
     pub api_provider: Arc<A::ApiProvider>,
+    pub submissions: Arc<SharedVector<InternalBidSubmission>>,
+    pub submission_results: Arc<SharedVector<SubmissionResultWithRef>>,
```
vladimir-ea: why do submission results need to be stored in a `SharedVector`? can the result not be contained in a spine message?

crates/relay/src/api/builder/submit_block.rs
```diff
@@ -32,41 +39,53 @@ impl<A: Api> BuilderApi<A> {
...
 
-        if let Err(err) = &res &&
-            err.should_report()
-        {
-            error!(%err)
-        }
+        let Some(future) = api.future_results.get(future_ix) else {
+            tracing::error!("failed to find future response in the shared vec");
+            return BuilderApiError::InternalError.into_response();
+        };
```
vladimir-ea, line 75: is this necessary? the value was pushed above @ line 57 - could just keep a reference or copy?

crates/relay/src/api/submission_results_fanout.rs
```diff
@@ -0,0 +1,74 @@
...
+impl Tile<HelixSpine> for SubmissionResultsFanOut {
+    fn loop_body(&mut self, adapter: &mut flux::spine::SpineAdapter<HelixSpine>) {
+        adapter.consume(|SubmissionResultIx { ix }, _producers| {
+            let Some(result) = self.submission_results.get(ix) else { return };
+            let SubmissionRef::Http(future_ix) = result.sub_ref else { return };
+            let Some(future) = self.future_results.get(future_ix) else { return };
+            future.set(ix);
+        });
+    }
+}
```
vladimir-ea: I think perhaps this tile is not needed. We could potentially simplify:
- create a 'produce result' function that:
- if there is http ref: 
-- looks up result future in the result futures shared vec
-- stores the result IN the future result struct 
-- calls 'wake()' on the stored future
-- stored future returns the Result
- if there is no http ref (so tcp)
-- produces the result message on a spine queue

crates/relay/src/auctioneer/submit_block.rs
```diff
@@ -79,48 +76,52 @@ impl<B: BidAdjustor> Context<B> {
...
+        adapter: &mut flux::spine::SpineAdapter<HelixSpine>,
+    ) -> bool {
         let Some(result) = &mut result.1 else {
-            return;
+            return false;
         };
 
-        let res_tx = result.res_tx.take();
+        if result.optimistic_version.is_optimistic() {
+            return false;
```
vladimir-ea: why this new return here? we also need to process sim results for optimistic submissions don't we?

crates/relay/src/auctioneer/types.rs
```diff
@@ -48,25 +49,40 @@ use crate::{
     housekeeper::PayloadAttributesUpdate,
 };
 
+pub struct SubmissionResultWithRef {
+    pub sub_ref: SubmissionRef,
+    pub result: Result<(), BuilderApiError>,
+}
```
vladimir-ea: turn this into a spine message?

crates/relay/src/auctioneer/worker.rs
```diff
@@ -297,36 +308,34 @@ impl SubWorker {
...
-
-        Ok((blinded_block, block_hash))
-    }
 }
 
 impl Tile<HelixSpine> for SubWorker {
-    fn loop_body(&mut self, _adapter: &mut flux::spine::SpineAdapter<HelixSpine>) {
+    fn loop_body(&mut self, adapter: &mut flux::spine::SpineAdapter<HelixSpine>) {
+        adapter.consume(|NewBidSubmissionIx { ix }, producers| {
+            if let Some(bid) = self.submissions.get(ix) &&
```
vladimir-ea: should log error for case where the submission is not found? or can that happen?

crates/relay/src/api/builder/api.rs
ninaiiad: no, because they're not copy (and not clone) since they store errors

crates/relay/src/auctioneer/types.rs
ninaiiad: this is not copy

crates/relay/src/api/builder/api.rs
vladimir-ea: they can be made copy? error becomes rsp code and message? I don't think we need to design around the 'not copy' nature of errors

crates/relay/src/auctioneer/worker.rs
ninaiiad: should never happen if we allocate enough for a single slot and the worker is not super behind (None means the buffer has wrapped around or was cleared manually - this we never do)

ninaiiad: have added the log too

crates/common/src/config.rs
```diff
@@ -98,6 +98,7 @@ impl RelayConfig {
                 sub_workers: vec![],
                 reg_workers: vec![],
                 tcp_bid_submissions_tile: 2,
+                submission_results_fanout: 3,
```
vladimir-ea: still needed?

crates/relay/src/api/builder/submit_block.rs
```diff
@@ -32,41 +37,48 @@ impl<A: Api> BuilderApi<A> {
...
 
-        res
+        let result = FutureBidSubmissionResult::wait(future).await;
```
vladimir-ea: should we have a timeout on this wait do you think? as a backstop in case

vladimir-ea (approved):
Looks good - a couple of minor comments

crates/relay/src/main.rs
```diff
@@ -143,6 +148,14 @@ async fn run(instance_id: String, config: RelayConfig, keypair: BlsKeypair) -> e
 
         let (top_bid_tx, _) = tokio::sync::broadcast::channel(100);
 
+        let submissions = Arc::new(SharedVector::<InternalBidSubmission>::with_capacity(
+            MAX_SUBMISSIONS_PER_SLOT,
+        ));
```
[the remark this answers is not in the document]
ninaiiad: not this vector, it's a ring buffer

crates/relay/src/auctioneer/worker.rs
```diff
@@ -294,39 +307,41 @@ impl SubWorker {
...
+                        producers,
+                        &self.future_results,
+                        bid.submission_ref,
+                        Err(e),
+                    );
+                }
+            } else {
+                tracing::error!("failed to find encoded bid submission for ix {}!", ix);
+            }
+        });
```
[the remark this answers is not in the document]
ninaiiad: there is flux telemetry inside consume function

PR #413, opened 2026-03-06
Deduplicate get_payload requests

vladimir-ea:
---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/413" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/relay/src/auctioneer/handle.rs
+++ b/crates/relay/src/auctioneer/handle.rs
@@ -11,7 +11,7 @@ use tokio::sync::oneshot::{self, Receiver};
 use tracing::trace;
 
 use crate::{
-    api::proposer::ProposerApiError,
+    api::proposer::{ProposerApiError, get_payload::ProposerApiVersion},
     auctioneer::types::{Event, GetHeaderResult, GetPayloadResult, RegWorkerJob},
     gossip::BroadcastPayloadParams,
 };
@@ -39,7 +39,7 @@ pub enum GetPayloadKind {
 pub struct AuctioneerHandle {
     auctioneer: crossbeam_channel::Sender<Event>,
     /// Dedup concurrent get_payload calls with the same block signature.
-    inflight_payloads: Arc<DashMap<SignatureKey, SharedPayloadFut>>,
+    inflight_payloads: Arc<DashMap<(ProposerApiVersion, SignatureKey), SharedPayloadFut>>,
 }
 
 impl AuctioneerHandle {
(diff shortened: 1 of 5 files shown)


crates/relay/src/main.rs
```diff
@@ -242,6 +242,7 @@ async fn run(instance_id: String, config: RelayConfig, keypair: BlsKeypair) -> e
                 auctioneer_core,
                 future_results,
                 decoded,
+                auctioneer_handle.clone(),
```
ninaiiad: nit: doesn't need a clone I think?

crates/relay/src/auctioneer/handle.rs
```diff
@@ -39,43 +62,68 @@ impl AuctioneerHandle {
...
-                    tracing::error!("failed to send get_payload to auctioneer");
-                    return Err(ChannelFull)
-                }
+    ) -> Result<GetPayloadKind, ChannelFull> {
+        let sig_key: SignatureKey = blinded_block.signature().serialize();
+
+        match self.inflight_payloads.entry((api_version, sig_key)) {
+            dashmap::mapref::entry::Entry::Occupied(entry) => {
+                trace!("dedup get_payload hit");
+                Ok(GetPayloadKind::Dedup(entry.get().clone()))
```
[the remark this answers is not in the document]
vladimir-ea: @0w3n-d is this an issue? should only positive outcomes be cached?

crates/relay/src/auctioneer/handle.rs
```diff
@@ -1,27 +1,50 @@
...
+    },
+    /// Duplicate caller — shared result, only returns the response.
+    Dedup(SharedPayloadFut),
+}
+
 #[derive(Clone)]
 pub struct AuctioneerHandle {
     auctioneer: crossbeam_channel::Sender<Event>,
+    /// Dedup concurrent get_payload calls with the same block signature.
+    inflight_payloads: Arc<DashMap<(ProposerApiVersion, SignatureKey), SharedPayloadFut>>,
```
0w3n-d, line 42: We might not need a map here because a slot can't have more than one valid signature. If we receive more than one valid signature it means the proposer has equivocated. They have committed to more than one block, which is against the rules of the protocol. We should only accept the first valid signature.

crates/relay/src/auctioneer/handle.rs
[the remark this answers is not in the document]
0w3n-d: In the past when we ran multiple relays per region and gossiped payloads and also had the v2 and v3 stuff, I would have said yes, we need to be able to recover. But now I'm not sure if there is a way to recover.

I suppose sometimes we might not actually serve the winning header, but may still have the block. Like say quasar wins on ultrasound and the proposer requests the payload from both titan and ultrasound. If the proposer is in NV and quasar sent the block to our relay in FR for some reason. Then the proposers request will fail the first time in NV but I think we will request the missing block async from FR? Then if FR gossips the block to NV a proposer retry in NV will succeed.

It's an edge case but likely happens from time to time. However, in that example NV would have gossiped the get_payload request to FR and FR would have processed the request and published the block to the beacon node so it isn't really necessary for to gossip the block the NV. It's just an extra redundancy.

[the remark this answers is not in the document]
vladimir-ea: I have changed the error caching so that it caches only signature verification errors

crates/relay/src/auctioneer/handle.rs
vladimir-ea: I still want to cache errors in case we get some bogus requests with bad signatures

vladimir-ea: although .. does that happen?

0w3n-d: It might. I'm not sure.

0w3n-d (approved):

PR #415, opened 2026-03-11
mv sim to tile and add ssz support to simulator

0w3n-d:
---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/415" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/relay/src/auctioneer/mod.rs
```diff
@@ -65,28 +67,33 @@ pub struct Auctioneer<B: BidAdjustor> {
...
         bid_sorter: BidSorter,
         local_cache: LocalCache,
         bid_adjustor: B,
-        event_tx: crossbeam_channel::Sender<Event>,
         event_rx: crossbeam_channel::Receiver<Event>,
         id: usize,
         future_results: Arc<SharedVector<FutureBidSubmissionResult>>,
         decoded: Arc<SharedVector<SubmissionDataWithSpan>>,
+        sim_inbound: Arc<SharedVector<SimInboundPayload>>,
+        sim_outbound: Arc<SharedVector<SimOutboundPayload>>,
```
vladimir-ea: outbound means outbound from the sim node? so inbound / outbound is from pov of the sim node? potentially confusing ..

crates/relay/src/bid_decoder/tile.rs
```diff
@@ -214,9 +227,20 @@ impl DecoderTile {
...
+        // For plain SSZ full submissions, capture the decompressed bytes so the
+        // auctioneer can forward them to the simulator without re-encoding.
+        let sim_bytes = if !is_dehydrated &&
+            !with_mergeable_data &&
+            matches!(header.encoding, crate::bid_decoder::Encoding::Ssz)
+        {
+            Some((body.clone(), SubmissionFormat::FullSsz))
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
vladimir-ea: looks like a legitimate issue?

crates/relay/src/auctioneer/types.rs
```diff
@@ -175,6 +177,9 @@ pub struct SubmissionData {
     pub version: SubmissionVersion,
     pub withdrawals_root: B256,
     pub trace: SubmissionTrace,
+    /// Decompressed bytes to forward verbatim to the simulator, avoiding re-encoding.
```
ninaiiad: is it important that these are decompressed already? also, titan sends bids without compression so for them we can save a reference into the dcache here instead of copying out the bytes

vladimir-ea:
Looks good - a comment on the overall design: could we not just forward all received bytes to the simulator nodes? 
i.e. round robin bytes from the dcache to the sim nodes (and run a decoder tile on the sim nodes?) and then handle responses only in the relay itself. This would potentially offload some processing? 

Maybe as a future iteration of this?

crates/relay/src/auctioneer/types.rs
vladimir-ea: Agreed - in the case where the bytes can be forwarded as is, a reference to the network bytes can be used?

0w3n-d: Yeah it's unfortunate that some submissions are compressed. I did initially have it send the network bytes direct to the simulator but then later changed my mind. The simulator had to perform the same full decode process as the relay which just felt like too big a change for one PR. It's still my goal though.

0w3n-d:
> Looks good - a comment on the overall design: could we not just forward all received bytes to the simulator nodes? i.e. round robin bytes from the dcache to the sim nodes (and run a decoder tile on the sim nodes?) and then handle responses only in the relay itself. This would potentially offload some processing?
> 
> Maybe as a future iteration of this?

Yeah that's what I initially implemented last week but then changed my mind. It just felt like too big a change to do in one go. I wanted to make sure that the existing path would still work as is with current infra. I want to get this merged and deployed. Test it. And move the remaining geths to reths in production and so on before rolling out full change.

crates/relay/src/bid_decoder/tile.rs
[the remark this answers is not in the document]
0w3n-d: yeah will fix

crates/relay/src/auctioneer/mod.rs
0w3n-d: Yeah I'll rename these

crates/relay/src/simulator/tile.rs
```diff
@@ -0,0 +1,604 @@
...
+        let new = self.simulators.iter().any(|s| s.can_simulate_light());
+        let prev = self.accept_optimistic.load(Ordering::Relaxed);
+        if new != prev {
+            warn!(prev, new, "changing accept_optimistic simulation status");
+        }
+        self.accept_optimistic.store(new, Ordering::Relaxed);
+    }
+
+    fn handle_sim_request(&mut self, req: crate::simulator::ValidationRequest, fast_track: bool) {
+        assert_eq!(req.bid_slot, self.last_bid_slot);
```
ninaiiad, line 183: should this be an assert? maybe just an error log? unless we expect this to have been filtered out earlier?

ninaiiad (approved):

crates/relay/src/simulator/tile.rs
0w3n-d: Yeah this came from the sim manager. I think it's never supposed to happen. Should be caught earlier unless there's a critical bug in the code.

Having said that I think it has happened on mainnet and we have a ticket open to look into it at some point.

PR #422, opened 2026-03-17
dcache update + collaborative decoders

ninaiiad:
---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/422" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/relay/src/bid_decoder/tile.rs
+++ b/crates/relay/src/bid_decoder/tile.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use alloy_primitives::B256;
 use flux::{
-    spine::SpineProducers,
+    spine::DCacheRead,
     tile::Tile,
     timing::{InternalMessage, Nanos},
 };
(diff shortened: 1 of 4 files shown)


crates/relay/src/bid_decoder/tile.rs
```diff
@@ -45,54 +45,58 @@ pub struct DecoderTile {
...
+                tracing::error!("dcache read failed");
+                send_submission_result(
+                    &mut adapter.producers,
+                    &self.future_results,
+                    new_bid.submission_ref,
+                    Err(BuilderApiError::InternalError),
+                );
+            }
+            DCacheRead::SpedPast => {
+                tracing::error!("submissions consumer got sped past");
```
vladimir-ea, line 96: is the `NewBidSubmission` message and the payload kept in the `DCache`? or is the `NewBidSubmission` message sent separately on a spine queue? Wondering if enough information could be included in the spine message to allow a response message to be sent in this case.

vladimir-ea (approved):

crates/relay/src/bid_decoder/tile.rs
ninaiiad: only the payload is in the dcache, so the message in the spine queue is sufficient for a response even if dcache payload is lost
this error, however, means that the message in the spine queue was overwritten too

PR #424, opened 2026-03-23
update flux + dcache integration

ninaiiad:
---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/424" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/relay/src/bid_decoder/tile.rs
+++ b/crates/relay/src/bid_decoder/tile.rs
@@ -2,11 +2,11 @@ use std::sync::Arc;
 
 use alloy_primitives::B256;
 use flux::{
-    spine::DCacheRead,
+    spine::{DCacheRead, SpineProducers},
     tile::Tile,
     timing::{InternalMessage, Nanos},
 };
-use flux_utils::{DCache, SharedVector};
+use flux_utils::SharedVector;
 use helix_common::{
     RelayConfig, SubmissionTrace,
     api::builder_api::MAX_PAYLOAD_LENGTH,
@@ -36,7 +36,6 @@ pub struct DecoderTile {
     chain_info: ChainInfo,
     cache: LocalCache,
     config: RelayConfig,
-    submissions: Arc<DCache>,
     decoded: Arc<SharedVector<SubmissionDataWithSpan>>,
     future_results: Arc<SharedVector<FutureBidSubmissionResult>>,
     buffer: Vec<u8>,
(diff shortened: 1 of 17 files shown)


crates/relay/src/spine/mod.rs
```diff
@@ -6,8 +6,8 @@ use flux::{communication::ShmemData, spine::SpineQueue, spine_derive::from_spine
 #[derive(Debug)]
 pub struct HelixSpine {
     pub tile_info: ShmemData<TileInfo>,
-
-    #[queue(size(2usize.pow(16)))]
+    // todo @nina
+    #[queue(size(2usize.pow(10)), mtu(1024))]
```
ninaiiad, line 10: will see if this can also be made configurable

crates/relay/src/bid_decoder/tile.rs
```diff
@@ -63,44 +61,44 @@ impl Tile<HelixSpine> for DecoderTile {
...
                         &self.future_results,
                         new_bid.submission_ref,
-                        Err(e),
+                        Err(BuilderApiError::InternalError),
                     );
                 }
+                DCacheRead::SpedPast => {
+                    tracing::error!("submissions consumer got sped past");
+                }
+                DCacheRead::Empty => {}
```
vladimir-ea, line 95: is this now a closure so that we can check that dcache is not updated while it is executing?

what happens if we execute this closure and then discover that the dcache entry is overwritten? we have already written into `self.decoded` and sent the `DecodedSubmission` message - should we wait to send the `DecodedSubmission` message until after we know that this has completed sucessfully?

crates/relay/src/spine/mod.rs
vladimir-ea: what impact does setting the `mtu` have?

ninaiiad: it's a theoretical max size of the dcache payload used to calculate dcache capacity
dcache and queue sizes must be aligned because it gives us the guarantee that in the worst case (where every message is of size mtu) if the dcache region laps then queue does too, so the seqlock version of a given queue slot a consumer checks will reflect that

crates/relay/src/bid_decoder/tile.rs
ninaiiad: not really
the first closure reads the dcache entry which may or may not be invalid
this second closure allows us to handle the result of that read - we will know if it's invalid and will not produce another message on some other spine queue or return a successful response to the user

crates/relay/src/spine/mod.rs
ninaiiad: https://github.com/gattaca-com/helix/blob/8211e4eb5ae35d8c7c2c6cccd680136aa5b32ac9/crates/relay/src/main.rs#L164

crates/relay/src/bid_decoder/tile.rs
vladimir-ea: so the dcache entry cannot be overwritten while that closure is being executed?

ninaiiad: no, it can always be overwritten, but whether or not it is only matters for the first closure

vladimir-ea: ok understood - so the cache entry is only read in the first closure - and that is the only place we have access to it?

vladimir-ea (approved):
looks good

crates/relay/src/bid_decoder/tile.rs
ninaiiad: yes

0w3n-d (approved):

PR #426, opened 2026-03-25
init data-gatherer

ninaiiad:
---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/426" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/relay/src/data_gatherer/s3.rs
```diff
@@ -0,0 +1,81 @@
...
+        buf.extend_from_slice(&header_len.to_le_bytes());
+        buf.extend_from_slice(header_slice);
+        buf.extend_from_slice(payload);
+        let bytes = buf.freeze();
+
+        self.pending.push((id, bytes));
+    }
+
+    pub async fn flush(&mut self) {
+        for (id, payload) in self.pending.drain(..) {
```
ninaiiad: not exactly sure blocking the runtime on this loop is going to be better performance-wise than a tokio-task per insert

0w3n-d: I've been thinking about a similar issue on the housekeeper side. I started down the path of removing tokio but it got needlessly complex, so I've parked it for now.

Instead I've been thinking something like this could work?

```
pub struct TileRt {
    rt: Runtime,
    tasks: Vec<Pin<Box<dyn Future<Output = ()>>>>,
}

impl TileRt {
    pub fn push(&mut self, fut: impl Future<Output = ()> + 'static) {
        self.tasks.push(Box::pin(fut));
    }

    pub fn poll(&mut self, timeout: Duration) {
        self.tasks.retain_mut(|fut| {
            match self.rt.block_on(timeout(timeout, fut.as_mut())) {
                Ok(_) => false,   // done, drop it
                Err(_) => true,   // still pending, keep it
            }
        });
    }
}
```

Instead of a global tokio with dedicated tokio threads. Each tile that has an unavoidable tokio dependency, gets it's own tokio runtime that it uses internally to drive async tokio tasks.

In the dial9 [blog post](https://tokio.rs/blog/2026-03-18-dial9l). They mention that one of the first things they noticed was tokio tasks bouncing between cores. They point out that "applications can benefit from a runtime-per-core architecture to avoid cache line bouncing". I thing that makes sense. In a tiling architecture like ours each tile should own it's own tokio runtime to maintain cache locality.

crates/relay/src/data_gatherer/tile.rs
```diff
@@ -0,0 +1,228 @@
...
+                32; /* parent_hash */
+                const BUILDER_PUBKEY_OFFSET: usize = BLOCK_HASH_OFFSET + 32; /* block_hash */
+
+                if buf.len() < BUILDER_PUBKEY_OFFSET + BlsPublicKeyBytes::len_bytes() {
+                    return None;
+                }
+
+                let (slot, block_hash, builder_pubkey) = unsafe {
+                    (
+                        core::ptr::read_unaligned(buf.as_ptr() as *const u64),
```
0w3n-d: Does this work? I thought ssz uses little endian.

ninaiiad: our machines would also be little-endian, but will make the read explicit

0w3n-d: Yeah fair point. You can leave as is.

crates/relay/src/data_gatherer/tile.rs
```diff
@@ -65,14 +70,10 @@ impl DataGatherer {
                 .map
                 .extract_if(|_, v| v.slot < new_slot)
                 .map(|(hash, info)| Self::make_row(self.instance_id.clone(), hash, info));
-            block_on(ch.publish(rows));
+            self.rt.block_on(ch.publish(rows));
```
0w3n-d: Should this no be a spawn now? Currently this would block the thread until we get a response back from clickouse. We might only have one clickhouse in say FR and this could be writing from TK.

If just spawned here it'll be driven forward by the block_on at the end of loop anyway.

0w3n-d (approved):

PR #434, opened 2026-04-14
Od/sysupdate

0w3n-d:
This PR refactors the housekeeper module. Making it a tile.
It's adds a http client to handle requests from the housekeeper to the beacon node without the need for tokio.
It replaces the RegWorker with a registration tile.
Moves the website out into it's own crate.

---

<a href="https://app.devin.ai/review/gattaca-com/helix/pull/434" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/relay/src/housekeeper/tile.rs
```diff
@@ -0,0 +1,495 @@
...
+                    Poll::Ready(Ok(resp)) => {
+                        let slot = resp.data.head_slot.as_u64();
+                        if best.as_ref().is_none_or(|(_, s)| slot > *s) {
+                            best = Some((idx, slot));
+                        }
+                    }
+                    Poll::Ready(Err(e)) => warn!(%e, idx, "sync status fetch failed"),
+                }
+            }
+            if self.sync_check_pending.is_empty() &&
```
ninaiiad, line 401: looks like we can lose an update if there're pending clients
shouldn't we update the best_idx once any best client is found?

crates/relay/src/housekeeper/chain_head.rs
```diff
@@ -0,0 +1,382 @@
...
+// );
+
+impl ChainHead {
+    pub fn new(chain_info: Arc<ChainInfo>) -> Self {
+        let current_slot = chain_info.current_slot();
+        Self {
+            slot: current_slot,
+            block: None,
+            state: ChainHeadState::Pending,
+            next_deadline: Instant::now() + slot_cutoff_timeout(&chain_info, current_slot),
```
ninaiiad: shouldn't the slot_cutoff_timeout use previous slot here? i.e. current_slot-1

crates/relay/src/housekeeper/tile.rs
```diff
@@ -0,0 +1,495 @@
...
+    fn on_new_slot(&mut self) {
+        let slot = self.chain_head.head();
+        info!(%slot, "new slot started");
+        self.pending_il = None;
+        self.il_consensus_rx = None;
+
+        if let Some(cfg) = &self.primev_config {
+            self.primev_builders_fetch = PrimevBuildersFetch::new(&self.http_client, cfg);
+            self.primev_validators_fetch =
+                PrimevValidatorsFetch::new(&self.http_client, cfg, self.duties.clone());
```
ninaiiad: on epoch boundary these would be previous epoch boundaries, right? we should call this after the new duties are fetched

crates/relay/src/bid_decoder/tile.rs
```diff
@@ -135,6 +132,19 @@ impl Tile<HelixSpine> for DecoderTile {
...
             },
         );
+
+        adapter.consume(|msg: SlotMsg, _producers| {
+            let Some(ev) = self.slot_events.get(msg.ix) else {
+                tracing::error!(?msg, "slot event not found");
+                return;
+            };
+            if self.current_slot < ev.bid_slot {
+                self.http_submissions.clear();
```
ninaiiad: given that this is shared, not sure how it would work if we try to clear this from multiple concurrent tiles, could maybe delete new slot data too - could we maybe somehow ensure to only call `clear` once?

crates/common/src/http/client.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn connect(&self, url: &Url) -> Result<(Transport, MioPoll), HttpClientError> {
+        let https = match url.scheme() {
+            "https" => true,
+            "http" => false,
+            s => return Err(HttpClientError::UnsupportedScheme(s.to_string())),
+        };
+        let host = url.host_str().unwrap();
+        let port = url.port_or_known_default().unwrap();
+
+        let addr = (host, port).to_socket_addrs()?.next().unwrap();
```
ninaiiad: is this unwrap safe?

crates/common/src/http/transport.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+        if let Err(e) = this.tls.process_new_packets() {
+            return Poll::Ready(Err(io::Error::new(io::ErrorKind::Other, e)));
+        }
+        let cap = buf.remaining();
+        if cap == 0 {
+            return Poll::Ready(Ok(()));
+        }
+        let mut tmp = vec![0u8; cap];
```
ninaiiad: could we maybe have a buffer in Self that we reuse across calls for this? or if it's small and known to be under a certain size could keep it on the stack

crates/common/src/http/transport.rs
```diff
@@ -0,0 +1,178 @@
...
+        cx: &mut Context<'_>,
+        mut buf: hyper::rt::ReadBufCursor<'_>,
+    ) -> Poll<io::Result<()>> {
+        match self.get_mut() {
+            Transport::Plain(tcp) => {
+                let cap = buf.remaining();
+                if cap == 0 {
+                    return Poll::Ready(Ok(()));
+                }
+                let mut tmp = vec![0u8; cap];
```
ninaiiad: same here

crates/relay/src/housekeeper/tile.rs
```diff
@@ -380,9 +397,7 @@ impl Tile<HelixSpine> for HousekeeperTile {
                     Poll::Ready(Err(e)) => warn!(%e, idx, "sync status fetch failed"),
                 }
             }
-            if self.sync_check_pending.is_empty() &&
-                let Some((best_idx, _)) = best
-            {
+            if let Some((best_idx, _)) = best {
```
ninaiiad: this is better but still local for the current iteration and the upcoming pending bests will overwrite this one, so maybe we'd want to store the best of the batch in self?

ninaiiad (changes requested):

crates/relay/src/bid_decoder/tile.rs
```diff
@@ -138,13 +149,16 @@ impl Tile<HelixSpine> for DecoderTile {
                 tracing::error!(?msg, "slot event not found");
                 return;
             };
-            if self.current_slot < ev.bid_slot {
+            let new_slot = ev.bid_slot.as_u64();
+            let old_slot = self.current_slot.load(Ordering::Relaxed);
+            if new_slot > old_slot &&
+                self.current_slot
```
ninaiiad: I still don't really like this design-wise - we're not really meant to clear this anyway, it's a ring buffer that wraps
also, this should work with both orderings as Relaxed

ninaiiad (approved):

PR #439, opened 2026-05-18
feat: builder demotion alert with repromote button

### LINKED ISSUE #437: Telegram Bot to handle demotions
When a builder is demoted the helix should send a telegram message that includes the following details:
link: e.g. beaconcha.in/slot/14282203
slot: e.g. 14282203
network: e.g. mainnet
region: <region>
builder_id: <id>
builder_pubkey: <pubkey>
block_hash: <hash>
simulation failed: <error message>

followed by a "repromote" button
the button when clicked should send a http request to a url embedded in the button.
_e.g. https://helix.xyz/relay/v2/data/promote?token=<base64 encoded random u128>

We also will need to create this endpoint. The endpoint will need access to this cached token that was generated when the telegram message was sent. If the token in the request matches we promote the corresponding builder pubkey.

Upon successful repromtion we should send another telegram message with a message like:
optimistic promotion successful for builder: <id>

tosynthegeek:
### Alert on demotion
When a builder is demoted due to block validation failure in `_get_payload`, a Telegram message is sent to all configured chat IDs containing:
- beaconcha.in link for the slot
- slot, network, region, builder pubkey, builder ID, block hash, reason
- Inline "🔁 Repromote" button linking to the promote endpoint with a one-shot UUID token

### Repromote endpoint
`GET /relay/v2/data/promote?token=<uuid>` on BuilderApi:
- Validates the token against the in-memory PromotionTokenCache (tokens are one-shot and consumed on use)
- Promotes the builder in LocalCache
- Fires `db_promote_builder` to update `builder_info` in the DB
- Sends a success Telegram alert

### AlertManager changes
- Added `PromotionTokenCache` (`Arc<DashMap<String, BlsPublicKeyBytes>>`) to the `Telegram` variant
- Added `send_demotion` for sending the alert with the inline keyboard
- Added `generate_token` and `consume_promotion_token` helpers
- `AlertManager` is now constructed once as `Arc<AlertManager>` and shared between `ProposerApi` and `BuilderApi`

Closes #437

### DIFF
--- a/crates/common/src/alerts.rs
+++ b/crates/common/src/alerts.rs
@@ -1,17 +1,44 @@
+use std::sync::Arc;
+
+use alloy_primitives::B256;
+use dashmap::DashMap;
+use helix_types::BlsPublicKeyBytes;
 use teloxide::{
     Bot,
     payloads::SendMessageSetters,
     prelude::Requester,
-    types::{ChatId, ParseMode},
+    types::{ChatId, InlineKeyboardButton, InlineKeyboardMarkup, ParseMode},
 };
 use tracing::error;
+use uuid::Uuid;
 
 use crate::RelayConfig;
+
+#[derive(Clone, Default)]
+pub struct PromotionTokenCache(Arc<DashMap<String, BlsPublicKeyBytes>>);
+
+impl PromotionTokenCache {
+    pub fn insert(&self, pubkey: BlsPublicKeyBytes) -> String {
+        let token = Uuid::new_v4().to_string();
+        self.0.insert(token.clone(), pubkey);
+        token
+    }
+
+    // Consumes token (one-shot — can't reuse)
+    pub fn consume(&self, token: &str) -> Option<BlsPublicKeyBytes> {
+        self.0.remove(token).map(|(_, pubkey)| pubkey)
+    }
+}
 
 #[derive(Clone)]
 pub enum AlertManager {
     Disabled,
-    Telegram { bot: Bot, chat_ids: Vec<ChatId> },
+    Telegram {
+        bot: Bot,
+        chat_ids: Vec<ChatId>,
+        promotion_tokens: PromotionTokenCache,
+        relay_url: String,
+    },
 }
 
 impl AlertManager {
@@ -23,6 +50,8 @@ impl AlertManager {
                 AlertManager::Telegram {
                     bot: Bot::new(&alerts.telegram_bot_token),
                     chat_ids: alerts.chat_ids.clone(),
+                    promotion_tokens: PromotionTokenCache::default(),
+                    relay_url: alerts.relay_url.clone(),
                 }
             }
             _ => AlertManager::Disabled,
@@ -32,7 +61,7 @@ impl AlertManager {
     pub fn send(&self, message: &str) {
         match self {
             AlertManager::Disabled => {}
-            AlertManager::Telegram { bot, chat_ids } => {
+            AlertManager::Telegram { bot, chat_ids, .. } => {
                 let msg = message.to_owned();
 
                 for chat_id in chat_ids {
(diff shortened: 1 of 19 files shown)


crates/common/src/alerts.rs
```diff
@@ -51,4 +74,76 @@ impl AlertManager {
             }
         }
     }
+
+    pub fn send_demotion(&self, message: &str, token: &str) {
+        let AlertManager::Telegram { bot, chat_ids, .. } = self else { return };
+
+        let url = format!("https://helix.xyz/relay/v2/data/promote?token={token}");
```
0w3n-d: I know I put `helix.xyz` in the ticket but it would be good to have this host part configurable.

crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -1860,6 +1868,24 @@ impl PostgresDatabaseService {
         Ok(())
     }
 
+    #[instrument(skip_all)]
+    pub async fn db_promote_builder(
+        &self,
+        builder_pub_key: &BlsPublicKeyBytes,
+    ) -> Result<(), DatabaseError> {
+        let mut client = self.high_priority_pool.get().await?;
+        let transaction = client.transaction().await?;
```
0w3n-d: This doesn't need to be a transaction as it's a single statement.

crates/common/src/local_cache.rs
```diff
@@ -179,6 +179,22 @@ impl LocalCache {
...
+        let Some(mut builder_info) = self.builder_info_cache.get_mut(builder_pub_key) else {
+            return false;
+        };
+
+        if builder_info.is_optimistic {
+            return false;
+        }
+
+        builder_info.is_optimistic = true;
+        builder_info.is_optimistic_for_regional_filtering = true;
```
0w3n-d: We shouldn't set `is_optimistic_for_regional_filtering` to true here.
Nor should it be set to false on demotion. The `is_optimistic` is the only dynamic flag and it along controls whether a builder is optimistic.

crates/relay/src/api/proposer/get_payload.rs
```diff
@@ -390,13 +391,30 @@ impl<A: Api> ProposerApi<A> {
...
+                        );
+                        let token = self_clone.alert_manager.generate_token(builder_pubkey);
                         let _ = self_clone.auctioneer_handle.send_event(Event::BuilderDemotion {
                             slot,
                             builder_pubkey,
                             block_hash,
                             reason,
                         });
+
+                        self_clone.alert_manager.send_demotion(&message, &token);
```
0w3n-d: The main path by which builders get demoted is actually a failed simulation. So if the alert is raised here those demotions will not be included.

But all demotions in the code call into `https://github.com/gattaca-com/helix/blob/develop/crates/database/src/handle.rs#L87`

There's already a discord alert in there. Just need to add this telegram alert in there too.

0w3n-d (changes requested):

crates/common/src/alerts.rs
tosynthegeek: Fixed — moved the host to `AlertsConfig` as `relay_url` field, now constructed from config in `AlertManager::from_relay_config`.

crates/database/src/postgres/postgres_db_service.rs
tosynthegeek: Fixed — removed the transaction, using a direct `client.execute` instead.

crates/common/src/local_cache.rs
tosynthegeek: Fixed — removed, only `is_optimistic` is toggled now.

crates/relay/src/api/proposer/get_payload.rs
tosynthegeek: Fixed — moved the alert into `db_demote_builder` on `DbHandle` so it fires for all demotion paths (block validation and simulation failure). Added `alert_manager` to `Context` so both `handle_builder_demotion` and `handle_simulation_result` can pass it through to `db_demote_builder` alongside `network` and `region` from config and `builder_id` from the builder info cache.

0w3n-d (approved):

PR #455, opened 2026-07-16
adds header ts for merged blocks

0w3n-d:


crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -191,6 +191,12 @@ impl BlockMerger {
         record_step("get_header", start_time.elapsed());
         trace!("fetched merged header");
 
+        let merged_bid = self.local_cache.get_merged_block(entry.bid.block_hash());
+        if is_mev_boost && let Some(mut merged_bid) = merged_bid {
+            merged_bid.trace.header_served_time_ns = Some(utcnow_ns());
+            self.local_cache.update_merged_block(merged_bid);
```
ninaiiad: why do we need to reinsert it?

ninaiiad (approved):

crates/relay/src/auctioneer/block_merger.rs
0w3n-d: This was because get_merged_block returned a clone.

I've pushed a commit to make it better and fix the lints.

PR #460, opened 2026-07-28
websocket headers stream draft

ninaiiad:


ltitanb (approved):

crates/common/src/config.rs
```diff
@@ -672,6 +681,22 @@ pub struct InclusionListConfig {
     pub relay_address: Address,
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
+pub struct HeaderStreamConfig {
+    /// How long to stream for, ending at the client's `X-Timeout-Ms` deadline.
+    #[serde(default = "default_u64::<500>")]
+    pub stream_for_ms: u64,
+    /// Interval between bid updates.
+    #[serde(default = "default_u64::<50>")]
```
gd-0: lets go with 5ms default

crates/common/src/config.rs
```diff
@@ -672,6 +681,22 @@ pub struct InclusionListConfig {
     pub relay_address: Address,
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
+pub struct HeaderStreamConfig {
+    /// How long to stream for, ending at the client's `X-Timeout-Ms` deadline.
+    #[serde(default = "default_u64::<500>")]
```
gd-0: think we can probably tighten this. say to 200-300? what's the max latency we see in practice?

crates/relay/src/api/proposer/header_stream.rs
```diff
@@ -0,0 +1,254 @@
...
+        trace!("starting call");
+
+        let ValidatedHeaderRequest {
+            ms_into_slot,
+            validation_complete_ns,
+            user_agent,
+            is_mev_boost,
+            timeout_ms,
+            ..
+        } = proposer_api.validate_header_request(&params, &headers, &terminating)?;
```
gd-0, line 64: we should add an additional measure to make sure that only the proposer can access this stream.

now that we can redesign the protocol we can add any restrictions we want. We can't get commit-boost to sign the message (cc @ltitanb) - but maybe we can attach a unique ID to the proposer registration http header (registration is signed, so we can guarantee that this ID only the validator has), and make sure we get commit-boost to attach this ID when sending a request to connect to the header stream.

We can then check that this ID matches the validator registrations ID for the validator that is proposing this specific slot. We should also ensure that they can only connect to the stream a single time per slot

gd-0 (changes requested):

crates/relay/src/api/proposer/header_stream.rs
ltitanb: Don't think this works, you can't sign over the ID and anyone can get a valid validator registration from the relay data api

crates/common/src/config.rs
ninaiiad: for the get_header http requests? tail latency is super bad for these, p99 can go up to 2s

ltitanb (approved):

vladimir-ea (approved):

crates/common/src/config.rs
```diff
@@ -672,6 +680,25 @@ pub struct InclusionListConfig {
...
+pub struct HeaderStreamConfig {
+    /// How long to stream for, ending at the client's `X-Timeout-Ms` deadline.
+    #[serde(default = "default_u64::<300>")]
+    pub stream_for_ms: u64,
+    /// Interval between bid updates.
+    #[serde(default = "default_u64::<5>")]
+    pub interval_ms: u64,
+    /// Keys allowed to open a stream. Empty allows every connection.
+    #[serde(default)]
+    pub api_keys: HashSet<String>,
```
gd-0: reminder to never use std::HashSet. always Fx :)

crates/relay/src/api/proposer/header_stream.rs
```diff
@@ -0,0 +1,357 @@
...
+    get_header::{ValidatedHeaderRequest, resign_builder_bid},
+};
+use crate::api::{Api, HEADER_API_KEY, HEADER_API_TOKEN, router::Terminating};
+
+/// Frame layout: `[kind: u8][fork: u8][ssz SignedBuilderBid]`.
+const FRAME_HEADER_LEN: usize = 2;
+const FRAME_KIND_BID: u8 = 1;
+
+impl<A: Api> ProposerApi<A> {
+    /// Opens a bid stream for the given slot, parent hash and public key. Validated exactly like
```
gd-0: " Validated exactly like" - this isn't fully true right? as we validate the api key for the stream and not for get_header

crates/relay/src/api/proposer/header_stream.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    let Some(api_key) = headers
+        .get(HEADER_API_KEY)
+        .or_else(|| headers.get(HEADER_API_TOKEN))
+        .and_then(|key| key.to_str().ok())
+    else {
+        return Err(ProposerApiError::InvalidApiKey);
+    };
+
+    if !config.api_keys.contains(api_key) {
```
gd-0: we need to make sure that the api key is for the validator that is proposing this slot. not just that its a a valid api key

crates/relay/src/api/proposer/header_stream.rs
```diff
@@ -0,0 +1,357 @@
...
+            .min((GET_HEADER_REQUEST_CUTOFF_MS as u64).saturating_sub(ms_into_slot)),
+    )
+}
+
+fn build_frame(fork: ForkName, bid: &SignedBuilderBid) -> Vec<u8> {
+    let mut frame = Vec::with_capacity(FRAME_HEADER_LEN + bid.ssz_bytes_len());
+    frame.push(FRAME_KIND_BID);
+    frame.push(fork_byte(fork));
+    bid.ssz_append(&mut frame);
+    frame
```
gd-0, line 290: pretty sure socket.send can take by reference? if so then we can just alloc the frame vec once at the start of the loop and reuse it

crates/relay/src/api/proposer/header_stream.rs
```diff
@@ -0,0 +1,357 @@
...
+    fn api_key_accepted_from_either_header() {
+        let config = config_with_key("secret");
+
+        for name in [HEADER_API_KEY, HEADER_API_TOKEN] {
+            assert!(check_api_key(&config, &headers_with(name, "secret")).is_ok());
+        }
+    }
+
+    #[test]
+    fn api_key_rejected() {
```
gd-0, line 336: doesn't this just check  a hashmap works? not sure we need a test for any of the key stuff tbh

crates/relay/src/api/proposer/header_stream.rs
```diff
@@ -0,0 +1,357 @@
...
+        Extension(Terminating(terminating)): Extension<Terminating>,
+        headers: HeaderMap,
+        Path(params): Path<GetHeaderParams>,
+        ws: WebSocketUpgrade,
+    ) -> Result<impl IntoResponse, ProposerApiError> {
+        trace!("starting call");
+
+        let config = &proposer_api.relay_config.header_stream;
+
+        check_api_key(config, &headers)?;
```
gd-0: Is this api key map a fixed config? we want this to dynamically update for any validator connecting.

Flow should be:
- validator generates their own api key 
- validator adds to their commit boost config
- commit boost now attaches this key to all requests
- if this key is added to a valid registration (see below for some ideas on extending this past just new timestamp and valid sig) then we attach this key to that validator 

IIRC there were some issues with some clients not updating the timestamp. so maybe we can use other checks to determine if the registration is real. e.g., by matching the IP to the last valid registration. I think this will allow us to add a load more validators than if we have a manual whitelist. If the connection fails because we don;'t think the api key is legit from commit boost. we should respond with an error and commit-boost should fallback to the standard get_header

gd-0 (changes requested):

crates/relay/src/api/proposer/header_stream.rs
ninaiiad: no, it doesn't, it requires Into<Bytes>, an owner reference

crates/common/src/local_cache.rs
```diff
@@ -315,38 +326,55 @@ impl LocalCache {
...
+        let fee_recipient_changed = existing.fee_recipient != new.fee_recipient;
+        let gas_limit_changed = existing.gas_limit != new.gas_limit;
+
+        let resigned =
+            (fee_recipient_changed || gas_limit_changed || existing.timestamp != new.timestamp) &&
+                new.timestamp >= existing.timestamp;
+
+        if !resigned && existing_entry.ip_addr.is_some_and(|last| Some(last) != ip_addr) {
+            return RegistrationUpdate::IpMismatch;
+        }
```
gd-0, line 355: should add a comment as to why we do this (to let api key through)

crates/common/src/local_cache.rs
```diff
@@ -315,38 +326,55 @@ impl LocalCache {
...
+
+        let fee_recipient_changed = existing.fee_recipient != new.fee_recipient;
+        let gas_limit_changed = existing.gas_limit != new.gas_limit;
+
+        let resigned =
+            (fee_recipient_changed || gas_limit_changed || existing.timestamp != new.timestamp) &&
+                new.timestamp >= existing.timestamp;
+
+        if !resigned && existing_entry.ip_addr.is_some_and(|last| Some(last) != ip_addr) {
+            return RegistrationUpdate::IpMismatch;
```
gd-0, line 354: think we will return IPMismatch for anything that just hasn't changed. might be a little misleading

crates/common/src/local_cache.rs
```diff
@@ -404,10 +432,12 @@ impl LocalCache {
 #[cfg(test)]
 mod tests {
```
gd-0, line 435: a load of AI tests added here. There is no reason to test logic as simple as we added IMO. Often good to get AI to write the tests and verify they pass. but we can then just remove them to keep bloat / mantainance down

crates/common/src/local_cache.rs
```diff
@@ -315,38 +326,55 @@ impl LocalCache {
...
+        let new = &registration.message;
+
+        let fee_recipient_changed = existing.fee_recipient != new.fee_recipient;
+        let gas_limit_changed = existing.gas_limit != new.gas_limit;
+
+        let resigned =
+            (fee_recipient_changed || gas_limit_changed || existing.timestamp != new.timestamp) &&
+                new.timestamp >= existing.timestamp;
+
+        if !resigned && existing_entry.ip_addr.is_some_and(|last| Some(last) != ip_addr) {
```
gd-0, line 353: so i think there is a small vulnerability here for the first time we fill in the api key. If someone knew about this and insta replayed everyone as we deployed they could take ownership of the IP and then own setting the api key. 

But think given we will lock in the API key on their first re-registration (each epoch) the chances of this are near 0 and so fine

crates/common/src/local_cache.rs
ninaiiad: it's && ?

gd-0: right, but if the validator legitimately moves they are going to get this really weird error that only really matters internally to us no? best to just return whatever we did before

ninaiiad: they don't get to see this error, they see 200 Ok like they used to (same reg exists = don't retry & no error)

gd-0 (approved):

PR #461, opened 2026-07-29
helix operator api wiring

vladimir-ea:


crates/database/src/postgres/postgres_db_service.rs
```diff
@@ -1825,6 +1828,51 @@ impl PostgresDatabaseService {
...
+                    SELECT DISTINCT ON (demotions.public_key)
+                        demotions.public_key,
+                        demotions.block_hash,
+                        demotions.demotion_time,
+                        demotions.reason,
+                        demotions.slot_number
+                    FROM demotions
+                    INNER JOIN builder_info
+                    ON demotions.public_key = builder_info.public_key
+                    WHERE builder_info.is_optimistic = FALSE
```
0w3n-d, line 1850: We should add `AND builder_info.collateral > 0` there are one or two builders that used to be optimistic but not any more and we've refunded the collateral. They will pop up every time this runs.

We could also think of capping it in some way? Like last 30 days or something? That would also avoid pulling in really old ones? Or maybe we should just delete those builders from the db.

vladimir-ea: what is an appropriate max age for a demotion? do we have builders with collateral who remain demoted for more than a day?

0w3n-d: Normally no. People are pretty quick to promote. It might be fine without the date check. It's only really if some smaller builder gets demoted and we miss it or forget about. But I suppose we'd want to know about that so maybe it's OK to get a new alert.

I guess the edge case I was thinking about is someone gets demoted on our relay at weekend, we don't act on it. They are promoted in other relays and then a few days later we restart a server and message to demote that builder is sent to all connected relays. But once the p2p goes live that shouldn't happen anymore because as long as some relay promotes we'll get the promote message. So maybe it's not a real issue?

0w3n-d (approved):

crates/database/src/postgres/postgres_db_service.rs
vladimir-ea: it should be fine - we would send out the demotion message, operators who had seen the subsequent promotion will ignore it and they will send us the latest demotion / promotion state for the builder

PR #462, opened 2026-07-29
adds fields for merged block data api

0w3n-d:


crates/database/src/postgres/postgres_db_row_parsing.rs
```diff
@@ -289,11 +289,21 @@ impl FromRow for MergedBlockResponse {
             original_value: parse_numeric_to_u256(
                 row.get::<&str, PostgresNumeric>("original_value"),
             ),
-            merged_value: parse_numeric_to_u256(row.get::<&str, PostgresNumeric>("merged_value")),
+            // `merged_value` is the DB column name; the API/Rust-side name is `proposer_value`.
+            proposer_value: parse_numeric_to_u256(row.get::<&str, PostgresNumeric>("merged_value")),
+            total_merged_value: parse_numeric_to_u256(
+                row.get::<&str, PostgresNumeric>("total_merged_value"),
```
vladimir-ea: row.get requires the column to have a non-null doesn't it? will this barf on old rows without the new columns?

ninaiiad (approved):

crates/database/src/postgres/postgres_db_row_parsing.rs
0w3n-d: No because the db migration has `NOT NULL DEFAULT 0`

vladimir-ea (approved):

PR #463, opened 2026-07-30
reth version bump

0w3n-d:


crates/relay/src/network/inclusion_lists/consensus.rs
```diff
@@ -57,7 +57,7 @@ pub(crate) fn compute_shared_inclusion_list(
         })
         .collect();
 
-    InclusionList { txs: txs.into() }
+    InclusionList { txs: txs.try_into().expect("inclusion list exceeds spec limit") }
```
vladimir-ea: should error rather than panic maybe if this data coming from outside?

crates/relay/src/housekeeper/inclusion_list_service.rs
```diff
@@ -71,7 +71,7 @@ impl IlFetchState {
             Poll::Pending => Poll::Pending,
             Poll::Ready(Ok(resp)) => {
                 let txs: Vec<Transaction> = resp.result.into_iter().map(Transaction).collect();
-                Poll::Ready(Some(InclusionList { txs: txs.into() }))
+                Poll::Ready(Some(InclusionList { txs: txs.try_into().expect("inclusion list exceeds spec limit") }))
```
vladimir-ea: error rather than panic?

vladimir-ea (approved):
one comment on error vs panic

crates/relay/src/network/inclusion_lists/consensus.rs
0w3n-d: Pushed a fix for this and a bunch of other places I found were we did something similar

PR #470, opened 2026-08-10
Fix address parsing in disallow list construction

canercidam:
It seems that the existing parsing logic was creating an empty list, making the provided list ineffective.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6074,6 +6074,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with 1.14.0",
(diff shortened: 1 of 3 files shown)


0w3n-d:
existing code works fine, doesn't create any intermediate list

canercidam:
@0w3n-d If you check the first commit, the unit test seems to break with `B256` because an empty list is constructed so existing code doesn't seem to work. And it starts to work again after switching the to `Address`.

0w3n-d (approved):

PR #471, opened 2026-08-11
add operator group

vladimir-ea:


crates/common/src/config.rs
```diff
@@ -715,6 +715,9 @@ pub struct OperatorConfig {
     pub operators: Vec<Operator>,
     #[serde(default)]
     pub mode: OperatorP2pMode,
+    /// Operator group or company.
+    #[serde(default)]
+    pub operator_group: Option<String>,
```
gd-0: `ArrayStr` ?
Think it's reasonable to have a max size here

0w3n-d (approved):

PR #473, opened 2026-08-11
fix(validation): resolve the forwarder code hash from the actual state

ckoopmann:
# Issue

The `ensure_payment` function reads the forwarder account state from the [BundleState](https://docs.rs/revm-database/41.0.0/src/revm_database/states/bundle_state.rs.html#395), which only contains entries for accounts that have changed over the course of the bundle. 

However in the course of the payment tx, the state of the forwarder smart contract itself doesn't change, hence `output.state.state.get(&PAYMENT_FORWARDER)` returns `None` and the payment validation usually fails (except for the unlikely case that the smart contracts account changed in the course of a separate tx).

Every payment routed through the `PAYMENT_FORWARDER` is rejected whenever validation actually has to inspect it — i.e. when the fee recipient spends in-block and the balance-delta shortcut misses. This demoted a fully valid Titan payment at mainnet slot 14968908.

# Fix

Read the state from `self.provider.latest()` instead


crates/simulator/src/validation/mod.rs
```diff
@@ -909,6 +919,93 @@ pub trait BlockSubmissionValidationApi {
...
+            bundle
+                .state
+                .get(&recipient)
+                .and_then(|account| account.info.as_ref())
+                .map(|info| info.balance),
+            Some(value),
+            "the recipient received the full payment"
+        );
+        assert!(
+            bundle.state.get(&PAYMENT_FORWARDER).is_none(),
```
ckoopmann: Note this test, mainly just proves that the bundles tate doesn't contain an entry for the payment forwarder even if you execute a payment tx. 
Probably not necessary to merge the test.

gd-0: yeah lets remove the test

gd-0 (approved):
lgtm

PR #474, opened 2026-08-12
operator p2p payload sharing

vladimir-ea:


crates/relay/src/api/proposer/get_payload.rs
```diff
@@ -533,18 +533,32 @@ impl<A: Api> ProposerApi<A> {
...
-            bid,
+            &bid,
         );
-        self.gossiper.broadcast_payload(params).await
+        self.gossiper.broadcast_payload(params).await;
+
+        if let Some(operator_pubsub) = self.operator_api.as_ref()
+            && let Err(e) = operator_pubsub
+                .send(helix_types::OperatorMessage::Payload(Payload {
+                    ts_ms: utcnow_ms(),
```
gd-0, line 552: can we avoid using ms and instead always use nanos? it's the same size and fetching a ts returns ns from the sys call so it doesn;'t add anything

crates/operator/src/lib.rs
```diff
@@ -231,6 +235,9 @@ pub fn spawn_operator_connection(
                                     builder_collateral.operator_group,
                                 );
                             }
+                            OperatorMessage::Payload(payload) => {
+                                payload_handler(payload);
```
gd-0, line 243: where can I find the impl for this? might be im just blind 😅

vladimir-ea: its in relay/main.rs - its the closure parameter passed passed inhere

crates/relay/src/api/proposer/get_payload.rs
vladimir-ea: can do - but in network messages nano precision is a bit meaningless b/c clock drift is typically at least millis

gd-0: Not super stronlg opinionated here. But IMO I think it's worth standardising all timestamps in helix to be Nanos, no matter what.

crates/operator/src/lib.rs
gd-0: ah - does `gossip.try_send()` just write to our local DB then? thought that broadcast to everyone else but the local?

vladimir-ea: > ah - does `gossip.try_send()` just write to our local DB then? thought that broadcast to everyone else but the local?

that `gossip` is the existing sender for the incoming gossip -> auctioneer channel - the auctioneer caches the gossiped payloads for use in get_payload responses.

vladimir-ea: ends up being handled here: https://github.com/gattaca-com/helix/blob/50c3075b060676cc8539beee531bdd9643db3679/crates/relay/src/api/builder/gossip.rs#L11

0w3n-d (approved):

PR #477, opened 2026-08-14
revoke bundle if latest_only was set and builder has dropped it

0w3n-d:


crates/relay/src/auctioneer/block_merger.rs
```diff
@@ -179,6 +184,16 @@ impl BlockMerger {
...
+        let included_order_ids_set: FxHashSet<B256> =
+            response.included_order_ids.iter().copied().collect();
+
+        if included_order_ids_set.iter().any(|id| self.revoked_order_ids.contains(id)) {
+            warn!(
+                block_hash = %response.execution_payload.block_hash,
+                "merged payload contains an order revoked since it was built, rejecting"
+            );
+            return Err(PayloadMergingError::ContainsRevokedOrder);
+        }
```
vladimir-ea: this is comparing the orders in the merged block against the latest revoked orders? I think it should be checking against orders revoked in / up to the base block? otherwise we get:
- send block / merged orders to merging builder
- - new orders + orders revoked
- merged block returns, fails due to revoked orders

if orders are revoked frequently enough then no merged block is ever valid - the check needs to be 'does it contain orders that were revoked before the merged block build started'

crates/relay/src/block_merging/tile.rs
```diff
@@ -448,16 +502,30 @@ impl BlockMergingTile {
...
-                        // connection; filter so the stored merged bid never
-                        // regresses.
-                        let best =
-                            slot.best_merged.entry(merged.base_block_hash).or_insert(U256::ZERO);
-                        if merged.proposer_value <= *best {
+                        // Reject if built before the merge builder learned of a revocation.
+                        if merged
+                            .included_order_ids
+                            .iter()
+                            .any(|id| slot.revoked_order_ids.contains(id))
```
vladimir-ea: what exactly is being checked here? the comment is I think wrong? should reject a block if it is builder AFTER the merge builder learned of the revocation?

vladimir-ea (approved):

PR #480, opened 2026-08-14
add slot locked disperser

0w3n-d:
Similar to the PaymentForwarder and needed for the same reason, but used the block merging rev share payment.

### DIFF
--- a/contracts/src/SlotLockedDisperser.huff
+++ b/contracts/src/SlotLockedDisperser.huff
@@ -1,0 +1,107 @@
+/// @title Slot-Locked Ether Disperser
+///
+/// CALLDATA LAYOUT:
+/// Offset 0..3:    [uint32 timestamp] (4 bytes)
+/// Offset 4..23:   Recipient 1        (20 bytes)
+/// Offset 24..55:  Amount 1           (32 bytes)
+/// Offset 56..75:  Recipient 2        (20 bytes)
+/// Offset 76..107: Amount 2           (32 bytes)
+/// ... repeating [20-byte address][32-byte value] ...
+///
+/// Revert reasons are encoded as standard Solidity custom errors so callers
+/// can decode them with `cast` / an ABI instead of a bare revert(0,0):
+///   error AlreadyEntered();
+///   error TimestampMismatch();
+///   error TransferFailed(address recipient);
+///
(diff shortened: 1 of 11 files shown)


contracts/src/SlotLockedDisperser.huff
```diff
@@ -0,0 +1,107 @@
+/// @title Slot-Locked Ether Disperser
+///
+/// CALLDATA LAYOUT:
+/// Offset 0..3:    [uint32 timestamp] (4 bytes)
+/// Offset 4..23:   Recipient 1        (20 bytes)
+/// Offset 24..55:  Amount 1           (32 bytes)
```
gd-0, line 6: do we really need 32 bytes for amount? its in ETH. think a u128 is > total eth in the universe

contracts/src/SlotLockedDisperser.huff
gd-0, line 1: can you ask claude to tone down the comments here? quite hard to read 😅

contracts/src/SlotLockedDisperser.huff
```diff
@@ -0,0 +1,107 @@
...
+    transfer_failed:
+        // stack: [offset, calldata_word] -- recover the recipient that just failed
+        dup1 calldataload push1 0x60 shr   // [recipient, offset, calldata_word]
+        [SEL_TRANSFER_FAILED] 0xe0 shl push0 mstore   // mem[0:4)  = selector
+        push1 0x04 mstore                              // mem[4:36) = recipient
+        push1 0x24 push0 revert                        // revert(0, 36)
+
+    done:
+        // Refund remaining balance to msg.sender
+        selfbalance iszero exit jumpi
```
gd-0, line 88: is this case ever possible?

contracts/src/SlotLockedDisperser.huff
```diff
@@ -0,0 +1,107 @@
...
+        push0 push0 push0 push0
+        selfbalance caller gas call
+        iszero refund_failed jumpi
+
+    exit:
+        // Release the lock before returning control to the caller.
+        push0 [LOCK_SLOT] sstore
+        stop
+
+    refund_failed:
```
gd-0, line 98: when is this realistically going to happen? aren't we paying this with an eoa?

PR #482, opened 2026-08-19
combined adjustable + mergeable submissions

ckoopmann:
# Issue

A submission can carry bid adjustment data or block merging data, but not both. The two are signaled independently (`WITH_ADJUSTMENTS` flag, `merge_type`), but no body encoding carries both extensions, and the decoder drops one of the two when both are signaled: the dehydrated mergeable path ignores the with-adjustments flag, and the with-adjustments path errors on `merge_type: mergeable`.

A builder submitting mergeable blocks therefore loses bid adjustment on those bids entirely.

# Solution

Add a combined body encoding — core fields ++ `bid_adjustment_data` ++ `merging_data` — selected by the existing signals (with-adjustments flag set and `merge_type: mergeable`), and have the decoder return both extensions. Existing single-extension wire formats are unchanged.

The body layout matches the combined submission format already accepted by the ultra sound relay, so builders can use a single encoder for both relays.

# Details

- new flat body types `SignedBidSubmissionWithAdjustmentsAndMergingData` and `DehydratedBidSubmissionFuluWithAdjustmentsAndMergingData` carrying both `bid_adjustment_data` and `merging_data` (core fields ++ bid_adjustment_data ++ merging_data)
- decoder dispatches on (with_adjustments, merge_type) instead of treating them as exclusive: dehydrated and default paths decode the combined body when the with-adjustments flag is set and merge_type is mergeable, returning both merging and adjustment data
- existing single-extension wire formats and AppendOnly / mark_all_txs_mergeable behavior unchanged
- tests: SSZ round-trips for both new types, flat byte-layout assertion, decoder dispatch tests for both paths

### DIFF
--- a/crates/common/src/decoder.rs
+++ b/crates/common/src/decoder.rs
@@ -8,9 +8,12 @@ use flate2::read::GzDecoder;
 use flux_profiler::timed;
 use helix_types::{
     BidAdjustmentData, BlockMergingData, Compression, DehydratedBidSubmission,
-    DehydratedBidSubmissionFuluWithAdjustments, DehydratedBidSubmissionFuluWithMergingData,
-    ForkName, ForkVersionDecode, MergeType, SignedBidSubmission,
-    SignedBidSubmissionWithAdjustments, SignedBidSubmissionWithMergingData, Submission,
+    DehydratedBidSubmissionFuluWithAdjustments,
+    DehydratedBidSubmissionFuluWithAdjustmentsAndMergingData,
+    DehydratedBidSubmissionFuluWithMergingData, ForkName, ForkVersionDecode, MergeType,
+    SignedBidSubmission, SignedBidSubmissionWithAdjustments,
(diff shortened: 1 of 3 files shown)


crates/common/src/decoder.rs
```diff
@@ -350,6 +365,13 @@ impl SubmissionDecoder {
         body: &[u8],
     ) -> Result<(Submission, Option<BlockMergingData>, Option<BidAdjustmentData>), DecoderError>
     {
+        if self.with_adjustments && self.merge_type == MergeType::Mergeable {
```
0w3n-d: I think this needed to move up into `decode_merge`. `decode_default` will not be called when MergeType::Mergeable.

ckoopmann: Good catch — moved into decode_merge and restored decode_default;

0w3n-d (approved):

Issue #489, opened 2026-08-21
Gloas (ePBS) builder-API support


0w3n-d:
## Status
Helix currently implements the Bellatrix→Fulu builder-API shape: `getHeader`, `registerValidator`, `getPayload`. Gloas (ePBS) replaces these with `getExecutionPayloadBid`, `submitBuilderPreferences`, and `submitSignedBeaconBlock` respectively — a different flow, not a version bump of the same one. This issue scopes what's needed for helix to support it.

## Why
Gloas (EIP-7732) removes the execution payload from the beacon block body entirely. The block carries a `SignedExecutionPayloadBid` directly, and the full payload is revealed *after* the block is proposed via a `SignedExecutionPayloadEnvelope` broadcast to the PTC — not returned synchronously to the proposer. There's no more "blinded block" concept. Based on the Gloas builder-API spec: [builder.md](https://github.com/ethereum/builder-specs/blob/main/specs/gloas/builder.md), [validator.md](https://github.com/ethereum/builder-specs/blob/main/specs/gloas/validator.md), merged via [ethereum/builder-specs#165](https://github.com/ethereum/builder-specs/pull/165).

**Resolved design question:** does helix need a gossip/libp2p stack for envelope publishing? No — a connected beacon node's own HTTP API bridges HTTP↔gossip for every Gloas primitive (`POST /eth/v1/beacon/execution_payload_envelopes` publishes to gossip on helix's behalf; `/eth/v1/events?topics=...` streams gossip observations back). helix stays HTTP-only.

**Resolved design question:** who holds the on-chain builder identity (deposit, `builder_index`, signing key) that signs `SignedExecutionPayloadBid`/`SignedExecutionPayloadEnvelope`? For now, helix signs under its own configured identity (reusing the relay's existing signing key; `builder_index` is a placeholder `RelayConfig` field until a real on-chain registration exists), the same centralizing-trust model it already uses for `SignedBuilderBid`. Supporting external builders submitting their own signed bids/envelopes is deferred (tracked separately, not yet a step).

[...]

PR #503, opened 2026-08-24
Construct and sign the Gloas SignedExecutionPayloadEnvelope, and broadcast it to beacon nodes

0w3n-d:
Issue: #489 (step 2 of 6)

Stacked on #501 — base branch is `od/gloas-spec-link-cleanup`, not `develop`. Retarget to `develop` once #501 merges.

## What this PR does
- `BeaconClient`/`MultiBeaconClient::publish_execution_payload_envelope`: POSTs a `SignedExecutionPayloadEnvelope` to a connected beacon node's `/eth/v1/beacon/execution_payload_envelopes`, mirroring `publish_block`'s fan-out-and-return-on-first-success.
- `construct_signed_envelope`: builds and signs the envelope from a proposer's signed block and a held payload, guarding against a missing held payload, a held-payload/bid block-hash mismatch, and a bid committing to a different `builder_index` than helix's configured identity.
- `GloasBuilderIdentity`: helix's own Gloas builder identity (builder_index + key), signing under `DOMAIN_BEACON_BUILDER` — not the off-chain domain `SignedBuilderBid` uses. Reuses the relay's existing signing key; `builder_index` is a new `RelayConfig` field, defaulting to 0 as a placeholder until helix has a real on-chain builder registration.
- Wires all of the above into `submitSignedBeaconBlock`: decodes the block, constructs and signs the envelope, and broadcasts it via `MultiBeaconClient`.
- `NoHeldPayloads`: the `GloasPayloadStore` implementation used until step 3 lands a real one. Every real request currently gets a `NoHeldPayloadForBlock` error, which is correct: nothing has held a payload yet.

No real held-payload store yet (step 3). No support for external builders submitting their own signed bids/envelopes (deferred, see the issue's resolved design question).

## Tests
Written before wiring, per the issue's tests-first convention:
- `BeaconClient::publish_execution_payload_envelope`: correct headers/body, 200/202/error-response handling (httpmock).
- `MultiBeaconClient::publish_execution_payload_envelope`: returns on first success, errors only if all clients fail.
- `construct_signed_envelope`: correct construction, signature round-trip verification, and all three guard conditions.

## Reviewer checklist

### DIFF
--- a/crates/common/src/config.rs
+++ b/crates/common/src/config.rs
@@ -80,6 +80,10 @@ pub struct RelayConfig {
     pub enable_flux_profiler: bool,
     #[serde(default)]
     pub operator_config: Option<OperatorConfig>,
+    /// This relay's on-chain Gloas (ePBS) builder_index. Placeholder until helix has a real
+    /// on-chain builder registration; signs under the relay's own key in the meantime.
+    #[serde(default)]
+    pub gloas_builder_index: u64,
 }
 
 #[derive(Serialize, Deserialize, Clone)]
@@ -136,6 +140,7 @@ impl RelayConfig {
             clickhouse: None,
             enable_flux_profiler: false,
             operator_config: None,
+            gloas_builder_index: 0,
         }
     }
 }
--- a/crates/relay/src/api/proposer/submit_signed_beacon_block.rs
+++ b/crates/relay/src/api/proposer/submit_signed_beacon_block.rs
@@ -5,7 +5,7 @@ use axum::{Extension, http::HeaderMap};
 use helix_common::{chain_info::ChainInfo, decoder::Encoding, utils::extract_request_id};
 use helix_types::{
     BlsKeypair, Domain, EthSpec, ExecutionPayloadEnvelope, ExecutionPayloadGloas,
-    ExecutionRequestsGloas, ForkName, MainnetEthSpec, SignedBeaconBlock, SignedBeaconBlockGloas,
+    ExecutionRequestsGloas, ForkName, MainnetEthSpec, SignedBeaconBlockGloas,
     SignedExecutionPayloadEnvelope, SignedRoot,
 };
 use hyper::StatusCode;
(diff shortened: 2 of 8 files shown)


crates/common/src/config.rs
```diff
@@ -80,6 +80,10 @@ pub struct RelayConfig {
     pub enable_flux_profiler: bool,
     #[serde(default)]
     pub operator_config: Option<OperatorConfig>,
+    /// This relay's on-chain Gloas (ePBS) builder_index. Placeholder until helix has a real
+    /// on-chain builder registration; signs under the relay's own key in the meantime.
+    #[serde(default)]
+    pub gloas_builder_index: u64,
```
ninaiiad, line 86: are we sure it will be better for the relay to have it's own builder identity as opposed to acting as a proxy?

crates/relay/src/api/proposer/submit_signed_beacon_block.rs
```diff
@@ -1,27 +1,118 @@
...
+pub struct HeldGloasPayload {
+    pub payload: ExecutionPayloadGloas,
+    pub execution_requests: ExecutionRequestsGloas,
+}
+
+/// Looks up and consumes the payload held for a bid's committed block hash. Must not return
+/// the same payload twice.
+pub trait GloasPayloadStore: Send + Sync {
+    fn take_held_payload(&self, block_hash: B256) -> Option<HeldGloasPayload>;
+}
```
ninaiiad, line 29: why is it a trait?

crates/common/src/config.rs
0w3n-d: I think we would want to support both. But to support the builder being the signer will require a lot more work.
In this first version I just want to get the minimum done to get the helix working on gloas.

For the builder to sign and helix to just proxy the bid I think we'll want to add new submission endpoints and so on.

crates/relay/src/api/proposer/submit_signed_beacon_block.rs
0w3n-d: Yeah I'll fix this. Shouldn't be a trait.

PR #509, opened 2026-08-25
Wire submitBuilderPreferences to per-proposer per-slot storage on the auctioneer

0w3n-d:
Issue: #489 (step 4 of 6)

Stacked on #507 — base branch is `od/gloas-step3-execution-payload-bid`. Retarget to `develop` once #501/#503/#507 merge.

## What this PR does
- `BuilderPreferencesStore`: a `(proposer_pubkey, slot) -> max_execution_payment` map living on `Context` (not the per-slot-recycled `SlotContext`), since a proposer can submit preferences up to an epoch ahead of the slot they apply to -- entries must survive multiple slot transitions until their own slot arrives or passes. Pruned incrementally in `Context::on_new_slot`.
- New `Event::SubmitBuilderPreferences`, handled identically across all three auctioneer states (mirrors `BuilderDemotion`'s existing all-states pattern) since storing a preference doesn't depend on what the auctioneer is currently doing.
- Rejects submissions for a slot that's already passed with the existing `RequestForPastSlot` error, reusing the same check the spec requires ("the builder MUST also reject...preferences whose auth.message.slot has already passed").
- The handler now calls through to the auctioneer instead of discarding the request after logging it.

Nothing reads from this store yet. `getExecutionPayloadBid` (step 3) never enforces `max_execution_payment` -- there's no real bid to enforce it against until step 5 lands real payload submission, so wiring the read side now would be dead code. That wiring is a small follow-up once step 5 lands.

## Tests
- `BuilderPreferencesStore` (pure, no `Context` needed): stores and looks up per proposer per slot, resubmission overwrites the previous value, and `on_new_slot` prunes only entries for slots that have passed.

## Reviewer checklist

### DIFF
--- a/crates/relay/src/auctioneer/builder_preferences.rs
+++ b/crates/relay/src/auctioneer/builder_preferences.rs
@@ -1,0 +1,94 @@
+use helix_types::BlsPublicKeyBytes;
+use rustc_hash::FxHashMap;
+
+/// Per-proposer-per-slot `max_execution_payment` preferences, submitted via
+/// `submitBuilderPreferences` up to an epoch ahead of the slot they apply to. Lives on `Context`
+/// (not `SlotContext`), since entries must survive across slot transitions until their own slot
+/// arrives or passes.
+#[derive(Default)]
+pub struct BuilderPreferencesStore {
+    by_proposer_slot: FxHashMap<(BlsPublicKeyBytes, u64), u64>,
+}
+
+impl BuilderPreferencesStore {
+    pub fn store(
+        &mut self,
+        proposer_pubkey: BlsPublicKeyBytes,
+        slot: u64,
+        max_execution_payment: u64,
+    ) {
+        self.by_proposer_slot.insert((proposer_pubkey, slot), max_execution_payment);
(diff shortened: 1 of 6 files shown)


crates/relay/src/auctioneer/builder_preferences.rs
```diff
@@ -0,0 +1,94 @@
...
+        proposer_pubkey: &BlsPublicKeyBytes,
+        slot: u64,
+    ) -> Option<u64> {
+        self.by_proposer_slot.get(&(*proposer_pubkey, slot)).copied()
+    }
+
+    /// Drops entries for slots that have already passed, so a proposer's own resubmissions
+    /// (or one that never proposes) don't grow this unboundedly.
+    pub fn on_new_slot(&mut self, bid_slot: u64) {
+        self.by_proposer_slot.retain(|(_, slot), _| *slot >= bid_slot);
```
ninaiiad, line 34: how many slots will we need to store here for each proposer? slots per epoch?

ninaiiad (approved):

Issue #527, opened 2026-08-27
helix-builder simulation role: ethrex-based block validation alongside helix-simulator


0w3n-d:
## Status

`helix-simulator` is the only block validator. It is a reth node extension:
`crates/simulator/src/main.rs` builds a reth `EthereumNode` and merges a `relay`
RPC namespace into it, and `crates/simulator/src/validation/mod.rs` does the
validation work in reth/revm types (`RecoveredBlock<Block>`,
`BlockExecutionOutput`, `CachedReads`, `StateProviderBox`).

`helix-builder` already runs an embedded ethrex node
(`crates/builder/src/node.rs`: store, blockchain, devp2p, Engine API) and merges
blocks on ethrex EVM state (`crates/builder/src/engine/`). Today it always runs
the merging role: `main.rs` requires `--merging.config` and a `RELAY_KEY`, then
starts the merge engine and the merging TCP server tile.

This issue gives `helix-builder` a second, independently selectable role:
**simulation**. In that role the same embedded node serves the relay's
validation interface -- SSZ `/validate` and `/validate_merged`, plus the
`relay_validateBuilderSubmissionV5` and `relay_validateMergedBuilderSubmissionV5`
JSON-RPC methods -- with an ethrex port of the reth validation pipeline. The
relay selects it purely by config (`simulators[].url` / `ssz_url`); no relay
change.

## Why

Client diversity on the validation path: a reth-only simulator makes a reth
consensus bug a relay-wide bug. The builder already runs ethrex, so validating
on the same client the merge builder builds with also removes the
reth-vs-ethrex disagreement risk between merging a block and validating it.

Folding it into `helix-builder` rather than a new crate reuses the embedded-node
wiring (`node.rs`, 520 lines: datadir, store, blockchain, p2p, Engine API, head
watcher, graceful shutdown) as-is, with no extraction refactor and no second
copy to keep in step with the pinned ethrex rev.

## Roles

Roles are derived from which config files are supplied -- no new `mode` flag,
and existing merging-only deployments keep working unchanged:

| `--merging.config` | `--sim.config` | Result |
| --- | --- | --- |
| set | absent | merging only (today's behavior) |
| absent | set | simulation only |
| set | set | both, sharing one node |
| absent | absent | startup error |

`--merging.config` becomes optional. `RELAY_KEY` is only required when the
merging role is active, so a simulation-only deployment does not need one. The
merge engine, the flux spine and the merging server tile only start for the
merging role.

## Affected surface

[...]

PR #532, opened 2026-08-28
Execute a submitted block and check it against its header

0w3n-d:
Issue: #527 (step 4 of 10)

Base branch: `od/builder-sim-validation-step3` (#531, step 3). Retarget as the
stack merges.

## What this PR does

`BlockValidator::execute` runs `validate_block_pre_execution`, executes the
block on its parent state through ethrex's `Evm`, then checks gas used, the
receipts root and logs bloom, the requests hash and the state root. `validate`
chains it onto `prepare`.

The store is read-only throughout. `apply_account_updates_batch` computes the
post-state root in memory without committing, and nothing calls
`Blockchain::add_block`.

No payment check, no blacklist. Two checks the reth simulator performs are
deliberately absent, both verified rather than assumed:

- `validate_block_body`. It compares the transactions and withdrawals roots
  against the body, but the conversion computes both roots from that body, so it
  cannot fail. The relay checks the transactions root itself.
- Block access lists. `ExecutionPayloadV3` carries no BAL hash, so Amsterdam
  needs a newer payload version before this is checkable at all.

Inclusion lists are now out of scope for this simulator entirely; the issue's
step 7 is struck out. A block that violates a submitted list passes here and
fails on the reth simulator.

## Tests

Written first and signed off before implementation. 8 new, 20 in the file:

- A valid block executes and yields receipts and account updates.
- **Validating does not store the block.** The latest block number stays at 0
  and the block is absent from the store afterwards. This guards the property
  that makes this a simulator and not a node.
- Tampered state root, gas used and receipts root are each rejected.
- Execution requests the block did not produce are rejected.
- A bad base fee is rejected before execution, proving the pre-execution checks
  actually run.
- A transaction with an unusable nonce fails execution.

One fixture change: `bid_trace` recomputes the block hash from the converted
block, so a payload-tampering test reaches execution instead of stopping at the
step 3 bid-trace check.

`just fmt-check`, `just test` and `cargo clippy --all-features --no-deps -- -D warnings`
are clean. 54 tests pass in `helix-builder`, up from 46.

## Reviewer checklist

### DIFF
--- a/crates/builder/src/validation/mod.rs
+++ b/crates/builder/src/validation/mod.rs
@@ -10,7 +10,15 @@ use alloy_rpc_types::{
     beacon::{relay::BidTrace, requests::ExecutionRequestsV4},
     engine::ExecutionPayloadV3,
 };
-use ethrex_common::types::{Block, BlockHeader};
+use ethrex_blockchain::{BlockchainType, new_evm, vm::StoreVmDatabase};
+use ethrex_common::{
+    types::{AccountUpdate, Block, BlockHeader, ELASTICITY_MULTIPLIER, Receipt},
+    validation::{
+        validate_block_pre_execution, validate_gas_used, validate_receipts_root_and_logs_bloom,
+        validate_requests_hash,
+    },
+};
+use ethrex_crypto::NativeCrypto;
 use ethrex_storage::Store;
 use tokio::sync::watch;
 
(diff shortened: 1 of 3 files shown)


crates/builder/src/validation/mod.rs
```diff
@@ -51,6 +66,60 @@ impl BlockValidator {
...
+            .map_err(|e| ValidationError::PostExecution(e.to_string()))?;
+
+        let account_updates =
+            vm.get_state_transitions().map_err(|e| ValidationError::Execution(e.to_string()))?;
+        let state_root = self
+            .store
+            .apply_account_updates_batch(block.header.parent_hash, &account_updates)
+            .map_err(|e| ValidationError::Store(e.to_string()))?
+            .ok_or(ValidationError::MissingParentState)?
+            .state_trie_hash;
```
vladimir-ea, line 111: would be interested to know what goes on here!

vladimir-ea (approved):

Issue #537, opened 2026-08-31
Sim tile hydration cache is fed too late, dropping ~632k simulations/day


0w3n-d:
## What's wrong

The simulator tile drops ~632k simulations per day (~88 per slot) because it
cannot resolve the transaction hash references in dehydrated submissions.

`crates/relay/src/simulator/tile.rs:340` logs
`hydration failed in sim tile e=unkown tx: index N, hash H`. Measured on
`titan_relay.log.2026-08-30` (mainnet, `relay-mainnet-1-aws-fr`): 632,589
occurrences, the third-largest log source on the relay.

Each failure means the submission is never simulated:

- Optimistic submissions were already sorted into the auction by
  `auctioneer/submit_block.rs:73` and the builder already received `Ok(())`.
  The bid stays eligible for delivery with no validation behind it.
- Non-optimistic builders receive `InternalError` for a submission that was
  well-formed.

## Suspected cause

The relay keeps two transaction caches with different fill rules, and the one
that gates admission is not the one that hydrates.

| | Cache | Filled by |
|---|---|---|
| Auctioneer | `HydrationCache` (per builder pubkey) | every accepted submission, via `submit_block.rs:289` |
| Sim tile | `SimHydrationCache` (global) | only submissions it actually hydrates |

`auctioneer/validation.rs:39` admits a dehydrated submission when
`can_hydrate` passes against the *auctioneer's* cache. The sim tile then
hydrates against its own cache, which is only populated when it hydrates a
submission for real: on `spawn_sim`, or via the eviction path at
`simulator/tile.rs:249-256`. So the sim tile's cache is a strict subset of the
one that gated admission.

The failure follows from ordering. A submission carrying full transactions is
queued while every simulator is busy. A later dehydrated submission
referencing those transactions is fast-tracked, and
`handle_task_response` drains `priority_requests` before `requests`
(`simulator/tile.rs:307`), so it reaches `spawn_sim` while the submission it
depends on is still queued and unhydrated.

The same problem was already solved in the merging tile.
`block_merging/tile.rs:872`:

```rust
/// Even when a submission is not forwarded, its full transactions and
/// blobs must enter the cache so later dehydrated submissions from the
/// same builder can resolve their references.
fn feed_cache(&mut self, submission: &Submission) {
```

The sim tile never got the same treatment. `HydrationCache::feed` exists;
`SimHydrationCache` has no `feed` at all.

Two smaller defects sit in the same code path:

[...]

PR #547, opened 2026-08-31
Feed the sim tile's hydration cache on arrival (#537 step 1)

0w3n-d:
Issue: #537 (step 1 of 2)

## What this PR does

`SimHydrationCache` was filled only by `hydrate()`, so a submission the sim
tile queued instead of simulating left no trace in the cache, and the next
submission to reference its transactions could not be hydrated. This feeds the
cache when a request arrives, before the spawn-or-queue branch, so arrival
order no longer decides whether a submission can be simulated.

On mainnet (`titan_relay.log.2026-08-30`) that failure dropped 632,589
simulations in a day, about 88 per slot. Optimistic bids were already sorted
into the auction and acknowledged, so they stayed eligible with no validation
behind them; non-optimistic builders got `InternalError` for well-formed
submissions.

`BlockMergingTile::feed_cache` already does exactly this, with a comment
explaining why. This mirrors it in the sim tile, and removes the
eviction-time hydrate that was a partial workaround for the same problem.

- It does not construct `BlockSimError::HydrationMiss`, which is defined in
  `crates/common/src/simulator.rs:92` but never built — hydration failures
  still report `RpcError`. That is #537 step 2, together with the `unkown tx`
  typo in `crates/types/src/hydration.rs`.
- It does not make `SimHydrationCache` per-builder. It stays keyed globally by
  an `FxHash` of a transaction's last 67 bytes, while `HydrationCache` is keyed
  per builder pubkey. The open question on #537 covers whether that should
  change.
- It does not touch the ordering itself: `handle_task_response` still drains
  `priority_requests` before `requests`. Feeding on arrival makes that
  ordering harmless rather than removing it.

## Tests

Tests were written first, in commit `b34f920e`, before any implementation.
`sim_cache_misses_when_the_earlier_submission_was_only_queued` passed against
the unfixed code and is kept as the control.

`crates/types/src/hydration.rs`:
- `sim_cache_misses_when_the_earlier_submission_was_only_queued` — the failure
  this fixes, and still the behaviour when nothing feeds the cache
- `sim_cache_feed_enables_later_hydration_without_building_a_payload` — after
  `feed`, the reference resolves and `tx_cache_hits == 1`
- `sim_cache_feed_is_idempotent` — retries re-send the same bid
- `sim_cache_feed_ignores_hash_references` — a submission of only references
  adds nothing

`crates/relay/src/simulator/tile.rs`:
- `feed_cache_fills_from_a_submission_that_is_not_simulated`
- `feed_cache_ignores_full_submissions`

Two notes for the reviewer. The tests cover `feed_cache` directly rather than
through `handle_sim_request`, because that takes `&mut HelixSpineProducers`,
which is not constructible in a unit test — the same trade-off the merging
tile's tests make, so the one-line call site rests on review. And the tests
needed four `pub` helpers in `hydration.rs`
(`full_tx_for_test`, `tx_cache_key`, `tx_hash_ref_for_test`,
`dehydrated_submission_with_txs_for_test`), because
`DehydratedBidSubmissionFulu`'s fields are private to that module and the
relay crate cannot otherwise build a submission with chosen transactions. Say
if you would rather they sat behind a `test-utils` feature.

`just fmt-check`, `cargo clippy --all-features --no-deps -- -D warnings` and
`just test` are all green locally (232 passed, 0 failed, 8 pre-existing
ignored).

## Reviewer checklist

### DIFF
--- a/crates/relay/src/simulator/tile.rs
+++ b/crates/relay/src/simulator/tile.rs
@@ -999,7 +999,8 @@ mod tests {
     use alloy_primitives::{Address, U256};
     use helix_types::{
         BlobWithMetadata, BlobsBundle, ExecutionPayload, ExecutionRequests, MergedBlockTrace,
-        TestRandom,
+        TestRandom, dehydrated_submission_with_txs_for_test, full_tx_for_test, tx_cache_key,
+        tx_hash_ref_for_test,
     };
     use rand::{SeedableRng, rngs::SmallRng};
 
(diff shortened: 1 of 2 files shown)


0w3n-d:
Correction to this PR's description: the claim that `BlockSimError::HydrationMiss` "is never constructed" is **wrong**. Please do not read it as licence to delete the variant.

It is constructed and load-bearing:

- `crates/relay/src/simulator/client.rs:148` maps the simulator's HTTP 424 to it
- `crates/relay/src/simulator/tile.rs:436` branches on it to retry with full uncompressed SSZ bytes
- `is_merge_builder_attributable` (`block_merging/tile.rs:398`) excludes it explicitly, so a hydration miss never disables block merging

My grep had been scoped to `crates/common/src/simulator.rs` alone.

The true statement is narrower, and it changes what #537 step 2 should do. `HydrationMiss` means **the simulator** could not hydrate — the 424 path, for the not-yet-implemented simulator-side cache described at `crates/simulator/src/ssz_server.rs:55`. The relay-side failure this PR fixes is a different condition, and it reports `RpcError` via `infra_error()`.

So step 2 must **not** simply reuse `HydrationMiss` for the relay-side case: that would land on the retry branch at `tile.rs:436`, which exists for the simulator-side miss. It needs its own variant.

Nothing in the code this PR changes is affected. The error is only in the description above.

Issue #563, opened 2026-09-01
Hydration failures and builder submission timeouts cannot be attributed from the logs


0w3n-d:
## Status

Two of the relay's highest-volume failures cannot be attributed to anything
from the logs, which blocks fixing either.

**Hydration failures.** `simulator/tile.rs:340` logs
`error!(%e, "hydration failed in sim tile")` and nothing else. Measured on
`relay-mainnet-1-aws-fr` over 19 minutes on 2026-09-01 (build
`a737fa6f`): 8,235 failures, roughly 624k/day, against 34.2M
`helix_hydration_cache_hits` — a ~1.7% miss rate. Only 389 distinct
`(index, hash)` pairs, so a small set of transactions fails repeatedly, in
bursts.

That is consistent with at least three different causes, each needing a
different fix:

- one builder's dehydration disagreeing with the relay about what the relay
  already holds,
- the per-slot cache clear in `on_new_slot` racing submissions that reference
  the previous slot's transactions,
- the submission carrying the full transaction never reaching the sim tile at
  all (dropped upstream, or lost to a ring overrun).

The log carries no builder pubkey, no slot and no block hash, so there is no
way to tell which. Related and probably sharing a root cause: 273
`cannot hydrate` submission rejections and 4,834 `bad order ref` merging
rejects in the same window.

**Builder submission timeouts.** `api/router.rs:146` fired 12,015 times in the
same window (~911k/day), every one of them on `/relay/v1/builder/blocks`. The
tower timeout is 5s while the handler's own wait is 3s, so a slow pipeline
should surface as `submit_block.rs:99` — that fired only 487 times. Roughly
96% of these requests spend more than 5s somewhere outside the inner wait, and
the builder receives a bare `408` with no logged reason.

The metrics rule out the obvious: `helix_submission_client_to_server_latency_us`
is 1.3 ms mean with all 111,373 samples under 5s, and body read, decode and
signature verification are about 1 ms each. `helix_submission_trace_latency_us`
shows `simulation` at 248 ms mean, 500 ms p50 and 5 s p99, which is the
standing suspect — but the timeout log records only `uri` and `request_id`, so
it cannot be correlated with body size or client.

## Why

Both are among the largest error sources on the relay, and neither can be
root-caused from what is logged today. This is the shortest path to fixing
them; writing speculative fixes first risks another deploy cycle spent
guessing.

## Affected surface

`crates/relay/src/simulator/tile.rs` — add submission identity to the
hydration failure.
[...]

Issue #542, opened 2026-08-31
Move expected-behaviour warns/errors into per-slot counters so ERROR means urgent


0w3n-d:
## Status

The relay emits ~4.0M WARN/ERROR lines per day on mainnet, of which about
99.7% report ordinary behaviour driven by an external actor — a builder, a
proposer, a beacon node, a peer relay, or a client. An `ERROR` currently
carries no signal that anything needs attention, so nobody can watch the log,
and a real fault is invisible in the volume.

Measured on `titan_relay.log.2026-08-30`, `relay-mainnet-1-aws-fr`, using the
call-site counter in `scripts/` (see Open questions). 39 of 205 call sites
account for all of it.

Eight tiles already have the structure this needs: `report_slot_stats` with a
per-slot `SlotStats` reset on transition, in the housekeeper, simulator,
bid decoder, bid sorter, auctioneer, top bid, data gatherer and block merging
tiles. The target state is that a quiet slot logs one stats line per tile and
nothing else, and every remaining `WARN`/`ERROR` means someone should look.

Two mechanical faults inflate the volume on top of the classification work:

**`#[tracing::instrument(..., err)]` defaults to ERROR.**
`api/proposer/get_payload.rs:137` and `:210` and
`api/proposer/register.rs:39` use bare `err`, so every `Err` return becomes an
ERROR event with no message, ~11.6k/day. Every sibling handler already uses
`err(level = tracing::Level::TRACE)`:
`get_header.rs:102`, `header_stream.rs:42`, `get_execution_payload_bid.rs:27`,
`submit_signed_beacon_block.rs:22`, `submit_builder_preferences.rs:21`. The
three outliers should match. This also double-logs: a past-slot request
produces both the `warn!("request for past slot")` at `get_payload.rs:283` and
an ERROR from the instrument event.

**Two log calls have no message at all.**
`api/builder/submit_block.rs:94` is `tracing::error!(err = result.error_msg.as_str())`
— 90,437/day with no builder pubkey, no block hash, no slot, and no message to
grep for. `common/utils.rs:161` is `error!("{message}")`.

## Why now

Two real bugs (#537, #538) were invisible in this log for weeks because their
symptoms looked like every other line. Making the log quiet is what makes the
next one visible.

## Behavior must not change

Only log levels, log sites and the contents of `report_slot_stats` lines
change. No control flow moves. Existing tests cover the behaviour around each
site; the new tests assert counter increments rather than log output.

## Affected surface

[...]

PR #565, opened 2026-09-01
Attribute hydration failures and builder submission timeouts (#563 step 1)

0w3n-d:
Issue: #563 (step 1 of 2)

## What this PR does

Neither of the relay's two largest remaining failures can be traced to a cause
from the logs. This adds the fields needed to root-cause them, and nothing
else.

**Hydration failures** — `simulator/tile.rs` logged only the error string.
Measured on `relay-mainnet-1-aws-fr` over 19 minutes on 2026-09-01 (build
`a737fa6f`): 8,235 failures, ~624k/day, against 34.2M
`helix_hydration_cache_hits` — a 1.7% miss rate, from only 389 distinct
`(index, hash)` pairs. That is consistent with three different causes needing
three different fixes: one builder's dehydration diverging from the relay,
the per-slot cache clear racing submissions, or the submission carrying the
full transaction never reaching the sim tile. Without a builder pubkey there
is no way to tell which. Now logs `builder_pubkey`, `slot` and `block_hash`.

**Builder submission timeouts** — `api/router.rs` fired 12,015 times in the
same window (~911k/day), every one on `/relay/v1/builder/blocks`, logging only
`uri` and `request_id`. Now also logs `content_length` and `user_agent`.

## A note on cost and on secrets

The hydration fields are read off the original submission inside the error
arm, not computed before the `hydrate` call that consumes the clone. The hit
path — 34M/day — pays nothing for them. My first draft copied the pubkey,
slot and hash unconditionally; this avoids that.

The router deliberately does **not** log the API key from `HEADER_API_KEY`.
It would identify the builder exactly, which is what we want, but it is a
credential and must not reach 12GB of retained logs. `user_agent` is the best
non-secret proxy available in that closure — mapping the key to a builder id
needs cache access the `HandleErrorLayer` does not have. Flagged as an open
question on #563.

- It does not fix either failure. Step 2 is to read a day of attributed data
  and open the real fixes. Writing the fix first risks another deploy cycle
  spent guessing which of the three causes is real.
- It does not count hydration failures per builder in the simulator's
  `report_slot_stats` line. That would answer "is it one builder?" without
  grepping and fits #542's direction, but it is more code and would slow this
  down. Left as an open question.
- It does not touch the 5s router timeout versus the handler's 3s inner wait,
  which is the likely reason 96% of these requests return a bare 408 rather
  than a diagnosable error. That is a behaviour change and wants its own
  issue.

## Tests

None. This changes log fields only — there is no behaviour to assert, and
asserting on `tracing` output would test the subscriber rather than the code.
Verification is the deployed log: after this ships, `hydration failed in sim
tile` should carry a builder pubkey, and the timeout should carry a body size.

Flagging that explicitly because `CLAUDE.md` says tests define correctness, and
I am consciously not writing any here. Say if you would rather I added the
per-builder counter instead, which would be testable.

`just fmt-check`, `cargo clippy --all-features --no-deps -- -D warnings` and
`just test` are green locally (250 passed, 0 failed, 8 pre-existing ignored).

## Reviewer checklist

### DIFF
--- a/crates/relay/src/api/router.rs
+++ b/crates/relay/src/api/router.rs
@@ -14,7 +14,7 @@ use axum::{
 use helix_common::{Route, RouterConfig, utils::extract_request_id};
 use hyper::{
     HeaderMap, Uri,
-    header::{CONTENT_LENGTH, USER_AGENT},
+    header::{CONTENT_LENGTH, HeaderName},
 };
 use tower::{BoxError, ServiceBuilder, timeout::TimeoutLayer};
 use tower_governor::{
@@ -38,6 +38,10 @@ use crate::{
 pub struct Terminating(pub Arc<AtomicBool>);
 #[derive(Clone)]
 pub struct KnownValidatorsLoaded(pub Arc<AtomicBool>);
+
+/// The relay sits behind a load balancer, so the socket peer is the balancer;
+/// this header carries the actual caller.
+const X_FORWARDED_FOR: HeaderName = HeaderName::from_static("x-forwarded-for");
 
 pub fn build_router<A: Api>(
     router_config: &mut RouterConfig,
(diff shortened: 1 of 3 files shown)


0w3n-d:
Updated in `f14f77f6`, and the "no tests" caveat in the description above no longer applies — there are two now.

**Per-builder counter added.** `LocalTelemetry` gains `hydration_failures: FxHashMap<BlsPublicKeyBytes, u32>`, incremented at the miss, reported in the existing `simulator slot stats` line as `hydration_failures_total` plus the worst five builders. It rides the existing `std::mem::take` in `report()`, so it resets per slot for free. Tests: `hydration_failures_are_counted_per_builder` and `hydration_failure_counts_reset_each_slot`.

**`user_agent` dropped**, as it tells us nothing.

**On using the builder pubkey there instead — it isn't possible at that layer, and it's worth knowing why.** The pubkey lives in the signed body. The API layer never decodes it: `submit_block` reads the body straight into a ring and constructs `NewBidSubmission { expected_pubkey: default, has_expected_pubkey: false }`, and decoding happens later in the bid decoder tile. The submit route carries no API key either — the only header it reads is `HEADER_SEND_TS` — so there is nothing to resolve against `api_key_cache`. And a router timeout means the body was never fully read, so even moving the log into the handler would not help for the 96% that fail before the inner wait.

So I used `source_ip` from `x-forwarded-for` instead. Unlike `user_agent` it does identify a builder in practice, since submission IPs are stable, and the relay is behind a load balancer so the socket peer is useless. Alongside `content_length`, which tests the "big body" hypothesis for the >5s spend.

If you would rather not log caller IPs, say so and I will drop it — `content_length` and `request_id` alone still make the timeouts joinable to any handler-side line for the same request.

Full gate green: 252 passed, 0 failed, 8 pre-existing ignored.

PR #566, opened 2026-09-01
dedicated TCP decoder lane + local-dev mode improvements

gd-0:
## Dedicated decoder lane for TCP submissions

TCP and HTTP submissions share one `to_decode` queue and one decoder pool. However, they have quite different workloads. Decode latency is dominated by sigvery, all TCP can bypass sigverify as it has the apikey. 

That means, if we share decoders for TCP they can be stuck behind http submissions getting sigverified.

This PR adds the configuration to have dedicated TCP only lanes.

Two types of lanes:
1. All: polls both TCP and http subs
2. TCP only: only polls TCP

Some numbers from running locally and sending  ~115 HTTP subs/s of signed 0.9 MB blocks + ~140 TCP subs/s from bid-submitter:
| | p50 | p90 | p99 | >1ms |
|---|---|---|---|---|
| shared decoders (develop) | 89–151µs | 431–477µs | 758–1419µs | 0.2–2.5% |
| TCP lane | 61µs | 342µs | 693µs | 0.19% |

## Local Dev Mode
Made some modifications to local-dev mode so we can run more seamlessly with a local builder:
- Duties without a registration get an unsigned stand-in 
- fee-recipient and zero-value checks skipped 
- unknown builders optimistic with unlimited collateral
-  a known api key admits any pubkey on TCP.

## Misc
1. is `cores.housekeeper` was `None` it would busy spin as a background tile floating around. We pinned in prod configs so this is just a footgun rather than a prod issue. Will now set min_loop_duration if its not pinned.
3. All tile priorities `OSDefault`. `High` failed anyway without CAP_SYS_NICE. It was also found to be counter-productive previously. So removing it entirely.
4. more buckets for some metrics
5. bumped the dcache max block size to 6MB rather than 2MB. 2mb is quite tight


crates/common/src/decoder.rs
```diff
@@ -211,7 +210,7 @@ impl SubmissionDecoder {
         }
     }
 
-    #[timed]
+    #[cfg_attr(feature = "profile", flux_profiler::timed)]
```
0w3n-d: When the profiler is disabled in the config this is basically a no op. It as like a 1 ns overhead.
I'd prefer to just leave it always in the build. I like to turn it on and off on production machines when investigating perf issues. This change would require deploying an alternative build just to see what's going on.

gd-0: sure - reversed

0w3n-d (approved):

PR #583, opened 2026-09-08
FIX: adds merge_data deny list

0w3n-d:
This allows us to prevent a builder from contributing to the merge order pool if we think there's a problem with what they are sending.


crates/common/src/config.rs
```diff
@@ -280,6 +280,10 @@ pub struct BlockMergingConfig {
     /// headers ahead of a broader rollout, without flipping the safety valve for everyone.
     #[serde(default)]
     pub serve_merged_headers_allowlist: HashSet<BlsPublicKeyBytes>,
+    /// Builder pubkeys whose merge data is dropped at decode time. Keeps a builder that submits
+    /// bad merge data out of the merge order pool, without stopping its normal submissions.
+    #[serde(default)]
+    pub merge_data_denylist: HashSet<BlsPublicKeyBytes>,
```
gd-0: std::HashSet!
WHere is the clippy denying std::HashSet ? :)

Should *always* use FxHashSet/ FxHashMap in place of std

crates/relay/src/bid_decoder/tile.rs
```diff
@@ -344,21 +346,29 @@ impl DecoderTile {
 
         trace!("sending to auctioneer");
 
+        let merge_data_denied =
+            config.block_merging_config.is_merge_data_denied(submission.builder_pubkey());
+        if merge_data_denied && merging_data.is_some() {
+            trace!("builder is on the merge data denylist, dropping its merge data");
+        }
```
gd-0: will this trace log *ever* be used? feel like trace level is a total waste

0w3n-d: Yeah fair point. Thinking of doing a big refactor PR later this week to clean out all the ai slot that has crept in.

vladimir-ea (approved):

gd-0 (approved):

PR #608, opened 2026-09-17
Add independent dehydration v2 and merging v2 TCP shapes

gd-0:
**TL;DR**

- Two new opt-in TCP shapes, selected by header flags: `MERGING_V2` (compact merge orders) and `DEHYDRATED_V2` (fixed-width tx and blob refs).
- Live e2e against this relay (v1 and v2 senders on the same blocks): a dehydrated send drops from 6.7 kB to 3.8 kB at p50, 10.1 kB to 5.0 kB mean; relay decode 16.7 µs to 4.3 µs p50.
- Per submitter-to-relay connection that is roughly 13 MB to 8 MB per slot, about 5.4 MB saved per slot or ~39 GB per day.
- Both v2 shapes are now the relay's internal representations; v1 converts into them once on decode, so v2 sends do no conversion.
- Every flag combination decodes through three generic containers, adjustments included; nothing changes for HTTP or unflagged TCP.
- Requires the builder PR gattaca-com/builder#3251; deploy this first, the builder flags default off.

**Why**

Every dehydrated bid resends its merge orders and one 12-byte reference per cached transaction. On live blocks those two are about 6.5 kB of a 10.3 kB send, and a relay connection sees ~1,400 sends per slot.

**Merging v2** (`BidSubmissionFlags::MERGING_V2`)

- Every order is a bundle over a contiguous tx range, `OrderV2 { start: u16, len: u8, flags: u8 }`, a fixed 4-byte item with no per-item SSZ offset.
- A tx order is a one-tx bundle with `ALL_REVERT = can_revert`; `LATEST_ONLY` is the other flag.
- Bundles whose txs mix revert, drop and neither carry an `OrderTxCodes` entry with a 2-bit code per tx; on live data 5% of bundles need one.
- `BlockMergingDataV2` is the relay's merging type; v1 merging data compacts into it once on decode.
- The merge tile builds the merge builder's refs straight from `OrderV2`; a one-tx order without codes or `LATEST_ONLY` still goes as a tx ref, so solo txs keep their tx-hash identity.

**Dehydration v2** (`BidSubmissionFlags::DEHYDRATED_V2`)

- `DehydratedBidSubmissionFulu` is now both the relay's dehydrated submission and the v2 wire shape: the payload carries only full txs and new blobs, and `tx_refs` / blob `refs` give block order as the existing FxHash cache keys, `0` meaning the next full item.
- A cached tx costs 8 bytes instead of 12, a cached blob 8 instead of 48.
- `withdrawals_ref` carries the FxHash of the slot's withdrawals once the relay has them (`0` = inline), so the ~700-byte list goes once per slot; the relay caches it per builder like txs and blobs, and the block hash still validates it.
- The v1 wire shape is `DehydratedBidSubmissionFuluV1`; it converts into the relay type in one pass on decode, and HTTP JSON still deserialises it.
- `hydrate`, `can_hydrate` and `feed` have one ref-based path with the same work as the old in-place loop.
- The relay keys cached blobs by the FxHash of the commitment for both versions; v1 senders still send commitments and the relay hashes them.

**Composition**

- A v2 body is one SSZ container `[submission][adjustments?][merging?]`, trailers present per the header flags.
- `WithAdjustments<S, A>`, `WithMergingData<S, M>` and `WithAdjustmentsAndMergingData<S, A, M>` are derived generic containers, so any submission shape pairs with any merging shape without a struct per combination.
- v2 bodies are SSZ only and merging v2 needs `Mergeable`; anything else returns `V2Unsupported` (400).
- `SubmissionDecoderParams` gains `dehydrated_v2` and `merging_v2`; it is SSZ-shipped to the simulator and validation servers, so relay and simulator deploy together.

**Numbers**

Merge orders, 13,518 live OVH-RBX candidate blocks: 126.5 orders per block, 97 of them single revertible txs, 30 bundles of average length 1.05.

| merging data encoding | bytes per block |
|---|---|
| current v1 `BlockMergingData` | 2,540 |
| union of variable orders with two `Vec<u8>` each | 1,325 |
| fixed 4-byte order + tx codes (this PR) | 552 |
| same without `ALL_REVERT` | 1,825 |

Whole send, 10,776 live blocks replayed through a per-slot hydration cache and encoded both ways:

| per unique block | v1 | v2 |
|---|---|---|
| whole dehydrated submission | 10,309 B | 6,380 B |
| cached tx refs (232 txs) | 2,788 B | 1,858 B |
| merging data | 3,725 B | 708 B |
| everything except new txs and blobs | 8,530 B | 4,601 B |

Per slot per submitter-relay connection, using the observed 851 unique blocks and 1,376 sends: v1 ≈ 8.8 MB first sends + 4.5 MB rebids = 13.2 MB; v2 ≈ 5.4 + 2.4 = 7.9 MB. New tx bodies (0.5 MB) and new blobs (0.9 MB) per slot are unchanged. The replay predates the 8-byte blob refs, so the v2 figure is slightly conservative.

**Live e2e** (relay from this branch, `is_local_dev`, no simulator; bid-submitter from gattaca-com/builder#3251 with one v1 and one v2 sender on the same OVH-RBX block stream, ~4 minutes, before the withdrawals key landed)

| | v1 | v2 |
|---|---|---|
| submissions | 28,861 | 38,677 |
| bytes per send, p50 / mean / p99 | 6,732 / 9,892 / 15,142 | 3,843 / 5,010 / 9,213 |
| relay decode µs, p50 / p99 | 16.7 / 65.5 | 4.3 / 26.8 |
| relay hydrate µs, p50 / p99 (one path) | 4.2 / 30.0 | |
| decode errors, hydration errors | 0, 0 | 0, 0 |

v1 decode now pays the one-pass conversion into the v2-shaped internal type; v2 decodes straight into it.


crates/types/src/hydration.rs
```diff
@@ -360,107 +471,130 @@ struct BlobItemFulu {
...
-                return false;
-            }
-        }
-
-        true
+        self.blobs_bundle.refs.len() <= max_blobs_per_block &&
+            (self.withdrawals_ref == 0 || withdrawals.contains_key(&self.withdrawals_ref)) &&
+            self.tx_refs.iter().all(|&r| r == NEXT_FULL_ITEM || txs.contains_key(&r)) &&
+            self.execution_payload.transactions.iter().all(|tx| tx.len() >= TX_KEY_SIZE) &&
+            self.blobs_bundle.refs.iter().all(|&r| r == NEXT_FULL_ITEM || blobs.contains_key(&r))
```
0w3n-d, line 510: I think if the original block was v1 and then converted some of these refs could be new commitments for blobs in new_items. This would lead to a cache miss and return false. The old code used to also check against the new_items.

crates/types/src/hydration.rs
```diff
@@ -177,6 +180,71 @@ pub struct DehydratedBidSubmissionFulu {
...
+    message: BidTrace,
+    execution_payload: ExecutionPayload,
+    blobs_bundle: DehydratedBlobsFuluV2,
+    execution_requests: Arc<ExecutionRequests>,
+    signature: BlsSignatureBytes,
+    tx_root: Option<B256>,
+    tx_refs: Vec<u64>,
+    /// Key of the slot's withdrawals if the relay has them, else 0 and
+    /// `execution_payload.withdrawals` is inline.
+    withdrawals_ref: u64,
```
0w3n-d, line 200: I think we check the withdrawals root before we hydrate the block. Dehydrated withdrawals might fail that check.

crates/types/src/hydration.rs
gd-0: Good catch, fixed in 6889ffa: `can_hydrate` also accepts a ref matching a `new_items` commitment hash. `hydrate` already inserted new items before resolving refs, so only the pre-check was wrong.

crates/types/src/hydration.rs
gd-0: Right, the decoder computes the root on the still-dehydrated payload. Fixed in 6889ffa: the cache stores the withdrawals with their tree hash root, and `validate_submission` uses that root for a keyed send in place of the decoder's empty-list root. The post-hydration block hash check still covers a wrong list.

0w3n-d (approved):

PR #609, opened 2026-09-17
relay: poll top-bid sockets with mio, encode each update once

gd-0:
`TopBidTile` fans top-bid updates out to builder websockets from a pinned flux tile. Every loop it also called `read()` on every connection to pick up pings and close frames. That is one `recv` syscall per connection per spin, almost all returning `EAGAIN`, and an update that lands mid-pass waits for the rest of the pass.

This PR registers each upgraded socket with a `mio::Poll` and replaces the read pass with one zero-timeout `epoll_wait`. Only sockets reported ready are read, and each is drained until `WouldBlock` because mio readiness is edge-triggered. Sockets are deregistered when dropped.

Each `TopBidUpdate` was also SSZ-encoded into a fresh `Vec` once per connection. It is now encoded once per precision (millis and nanos) on first use and the same `Bytes` handle is sent to every connection. tungstenite owns the payload as `Bytes`, so the per-connection cost is a refcount clone.

The 10s ping timer moves from `Nanos` (wall clock) to `flux::timing::Instant` (rdtsc).

Send semantics are unchanged: a non-blocking send failure still drops the connection rather than stalling the fan-out.

`mio` becomes a direct dependency of `helix-relay` (`os-poll`, `os-ext`). It was already in the tree via flux-network.

Cost of the read pass per loop, measured on one pinned core (Linux 7.1, loopback sockets with no data):

| Connections | N × `recv` returning EAGAIN | one `epoll_wait(0)` |
|---|---|---|
| 1 | 171 ns | 58 ns |
| 8 | 1518 ns | 58 ns |
| 32 | 6068 ns | 52 ns |

`relay-mainnet-b-aws-nv` carries 33 top-bid connections, so the develop tile spends about 6 µs of every loop in empty reads before it checks the queue again. A local end-to-end run with one submitter and 8 probe clients showed no measurable difference between develop and this branch at that connection count; the effect is proportional to connections.


gd-0:
didn't make much of a difference. so not worth the complexity atm

